PR #1, opened 2018-12-25
Add initial implement of the resize controller

mlmhl:
This first PR contains:

1. Initial implement of the resize controller, which takes on most of the controller side volume resize work, such as check if a PVC requested resize operation, execute the resize operation, mark the resizing operation as finished, etc.
2. An example `Resizer`, which performs the real resize operation on specific storage types. Currently it does nothing and just returns as resizing succeeded.


cmd/csi-resizer/main.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/kubernetes-csi/external-resizer/pkg/util"
+
+	"k8s.io/klog"
+)
+
+var (
+	master       = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	kubeConfig   = flag.String("kubeconfig", "", "Absolute path to the kubeconfig")
+	resyncPeriod = flag.Duration("resync-period", time.Minute*10, "Resync period for cache")
+	workers      = flag.Int("workers", 10, "Concurrency to process multi resize requests")
```
gnufied: rename `multi` to `multiple` ?

cmd/csi-resizer/main.go
```diff
@@ -0,0 +1,85 @@
...
+			"before it is replaced by another candidate. This is only applicable if leader "+
+			"election is enabled.")
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+
+	resizerName := "csi/example-resizer"
```
gnufied: rename this?

pkg/controller/controller.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			klog.V(3).Infof("PVC %s/%s is deleted, no need to process it", namespace, name)
+			return nil
+		}
+		klog.Errorf("Get PVC %s/%s failed: %v", namespace, name, err)
+		return err
+	}
+
+	if !ctrl.pvcNeedResize(pvc) {
```
gnufied: I think just checking PVC alone here is not enough to prevent double processing of PVCs. For example - if user edits PVC and requests new size and while this resize is in-progress some other changes are done to same PVC. All such changes would trigger a resize.

gnufied: In some cases it may be unavoidable to double process same PVC but for most part I think we should avoid reconsidering a PVC which is already being resized. And that is why I think it may be worth comparing old vs new PVC size in update event.

pkg/controller/controller.go
```diff
@@ -0,0 +1,388 @@
...
+
+	return ctrl
+}
+
+func (ctrl *resizeController) addPVC(obj interface{}) {
+	objKey, err := getPVCKey(obj)
+	if err != nil {
+		return
+	}
+	ctrl.claimQueue.Add(objKey)
```
gnufied: hmm, do we really need to consider a PVC for resizing when it gets added/created for the first time?  I know it will be dropped by checks we have in place but it would cause unnecessary churn in controller loop.

pkg/resizer/trivial_resizer.go
```diff
@@ -0,0 +1,37 @@
...
+
+type trivialResizer struct{}
+
+func (r *trivialResizer) CanSupport(pv *v1.PersistentVolume) bool {
+	return true
+}
+
+func (r *trivialResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quantity) (newSize resource.Quantity, fsResizeRequired bool, err error) {
+	return requestSize, false, nil
+}
```
gnufied: so this is a placeholder file that will eventually call out to CSI plugin right?

cmd/csi-resizer/main.go
mlmhl: Done

cmd/csi-resizer/main.go
mlmhl: I will set `resizerName` to CSI driver's name after we implementing the CSI plugin code. Currently just set it to an example name temporarily to make the controller work.

pkg/resizer/trivial_resizer.go
mlmhl: Yeah, I intend to implement the CSI plugin in subsequent PRs.

pkg/controller/controller.go
mlmhl: This is to deal with a rare situation:
1. `external-resizer` stopped, maybe stopped manually  by administrator, or just due to an unexpected panic
2. User edits a PVC and requests new size
3. `external-resizer` started.

After `external-resizer` started, it will only receive an `Added` event of this PVC from the `Informer`, without any `Updated` events. So if we just simply ignore the `Added` events, this resize request will lost.

pkg/controller/controller.go
mlmhl: Sorry I missed the check in the update event handler when adjusting code structure.

mlmhl: Added the check in update event handler and put the pvc into queue only if `newSize.Cmp(oldSize) > 0`, PTAL.

gnufied:
Lets merge this. I will open some github issues to ensure that we do not miss items.

cmd/csi-resizer/main.go
```diff
@@ -0,0 +1,85 @@
...
+			"slot. This is effectively the maximum duration that a leader can be stopped "+
+			"before it is replaced by another candidate. This is only applicable if leader "+
+			"election is enabled.")
+	leaderElectionRenewDeadLine = flag.Duration("leader-election-renew-deadline", time.Second*10,
+		"The duration that non-leader candidates will wait after observing a leadership "+
+			"renewal until attempting to acquire leadership of a led but unrenewed leader "+
+			"slot. This is effectively the maximum duration that a leader can be stopped "+
+			"before it is replaced by another candidate. This is only applicable if leader "+
+			"election is enabled.")
+)
```
gnufied: Is there any known pattern for tunable patterns to expose for leader election? The external-attacher seems to expose far smaller number of tunables. cc @msau42

cmd/csi-resizer/main.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/kubernetes-csi/external-resizer/pkg/util"
+
+	"k8s.io/klog"
+)
+
+var (
+	master       = flag.String("master", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	kubeConfig   = flag.String("kubeconfig", "", "Absolute path to the kubeconfig")
+	resyncPeriod = flag.Duration("resync-period", time.Minute*10, "Resync period for cache")
+	workers      = flag.Int("workers", 10, "Concurrency to process multiple resize requests")
```
gnufied: We need a value of controlling csi timeouts. I filed that as a separate issue.

cmd/csi-resizer/main.go
mlmhl: External-attacher doesn't expose these three durations: `leaderElectionRetryPeriod`, `leaderElectionLeaseDuration`, and `leaderElectionRenewDeadLine`, it just use the default values:
https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/leader.go#L37

I make these args configurable because the core kube-controller-manager does so: https://github.com/kubernetes/kubernetes/blob/35b17dd6d08589bc6182d8038be1cf86d03c6ab0/cmd/controller-manager/app/options/generic.go#L81. However, users do not need to set these args in most cases, because we have provided appropriate default values.

I'm not sure which one is better, do we have any restrictions or rules for leader election?

cmd/csi-resizer/main.go
mlmhl: Thanks! I will add CSI related implement in next PR.

Issue #4, opened 2019-01-07
Ensure resizer handles double processing of PVCs without problems.


gnufied:
xref - https://github.com/kubernetes/kubernetes/pull/71611

mlmhl:
It seems that the race condition mentioned in https://github.com/kubernetes/kubernetes/issues/71470#issuecomment-442248513 does not exist in external-resizer. The in-tree expand controller uses a custom [sync mechanism](https://github.com/kubernetes/kubernetes/blob/10bb353a95c53831b9131e8c4cc539c87f34a0d6/pkg/controller/volume/expand/pvc_populator.go#L84) to resync pvcs, which is the root of the problem. But in external-resizer, we uses `SharedInformer`'s [resync mechanism](https://github.com/kubernetes-csi/external-resizer/blob/7373866c8a4850da71cab7ecb7e803beee1dd1ff/pkg/controller/controller.go#L97). With this mechanism,  `SharedInformer` will send us `Update` events periodically, the old object and new object are the same in these events. So this [judgement](https://github.com/kubernetes-csi/external-resizer/blob/7373866c8a4850da71cab7ecb7e803beee1dd1ff/pkg/controller/controller.go#L132) inside the update handler will intercept these events.

But when I review the code, I found that this mechanism raises another race condition, consider the following example:

1. User requests size 1Gi -> 2Gi
2. Resizer resizes the volume and adds a `PersistentVolumeClaimFileSystemResizePending` condition
3. Kubelet starts to resize file system
4. User requests size 2Gi -> 3Gi
5. Resizer resizes the volume and updates the `PersistentVolumeClaimFileSystemResizePending` condition.
6. Kubelet finishes the first resize request and clear the conditions.
7. Kubelet performs the second resize request.

Although second resize request can be finished in step 7, but the PVC condition is not right during step 6 and step 7(wanted `PersistentVolumeClaimFileSystemResizePending`, got empty), because we won't process the PVC again(due to the judgement mentioned above).

To resolve this problem, I think we need to remove the judgement mentioned above and add a version check mechanism like https://github.com/kubernetes/kubernetes/pull/71611. @gnufied Do you have any other good ideas?

gnufied:
Yeah I agree, I think we do need a resourceVersion check added in https://github.com/kubernetes/kubernetes/pull/71611 to external resizer as well.

mlmhl:
@gnufied As the PR #6 is already merged, could we close this issue?

PR #6, opened 2019-01-10
avoid concurrent processing of same PVC

mlmhl:
This PR try to handle double processing of PVCs correctly.

The resize controller will receive PVC Update events periodically due to `SharedInformer`'s resync mechanism, when handling these events, we need to check PVC's `ResourceVersion` to avoid race condition between controller and kubelet.

xref: https://github.com/kubernetes-csi/external-resizer/issues/4

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -24,7 +24,7 @@ import (
 	"github.com/kubernetes-csi/external-resizer/pkg/resizer"
 	"github.com/kubernetes-csi/external-resizer/pkg/util"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


gnufied:
The PR looks okay but I am thinking if there is a better way of handling this. Lets not merge this for now until we consider other approaches. This PR is not blocking us from doing rest of the external resizer work, so we can merge this later as well.

gnufied:
I have been thinking about this and I think rather than doing a previous version check etc, we should *always* default to doing a resourceVersion check while patching PVC and PV in this PR. This would solve majority of duplicate processing problems and I think it should be safe enough. @mlmhl what do you think? Can you update this PR to do that?

mlmhl:
@gnufied Agreed. And the PR is already updated, PTAL :)

pkg/controller/controller.go
```diff
@@ -392,15 +413,56 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
+
+func (ctrl *resizeController) patchPVCStatus(pvc, newPVC *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
+	key, err := getPVCKey(pvc)
+	if err != nil {
+		return nil, fmt.Errorf("generate unique key failed: %v", err)
+	}
+	// If pvc's ResourceVersion is equal to last processed ResourceVersion, it means this is a resync event,
+	// so we need to check PVC's version when patch its status to avoid double process on same PVC.
+	// More details see: https://github.com/kubernetes-csi/external-resizer/issues/4.
+	needVersionCheck := ctrl.lastProcessedPVCVersions.Get(key) == pvc.ResourceVersion
```
gnufied: I meant, lets not do all this stuff. Just add resource version to PVC and PV patch unconditionally. Let the patch operation fail and retry if we had stale object.

mlmhl: Oh, got it. Updated.

PR #7, opened 2019-01-30
add implement for CSI plugin

mlmhl:
This PR adds an initial implement of CSI resize plugin. Currently we can test this by the csi-test tool with PR https://github.com/kubernetes-csi/csi-test/pull/160


pkg/resizer/csi_resizer.go
```diff
@@ -0,0 +1,162 @@
...
+		return oldSize, false, errors.New("not a CSI volume")
+	}
+	volumeID := source.VolumeHandle
+	if len(volumeID) == 0 {
+		return oldSize, false, errors.New("empty volume handle")
+	}
+
+	ctx, cancel := timeoutCtx(r.timeout)
+	defer cancel()
+	// TODO: Get secrets used for expansion operation.
```
rootfs: you can get the secret here and remove the TODO :)

mlmhl: I'm not sure how to get these secrets. The external-attacher got publish/stage secrets from the PV object directly, and the external-snapshotter got snapshot secrets from `VolumeSnapshotClass.Parameters`. So which way should we adopt? Add a `ControllerExpandSecretRef` field to `CSIPersistentVolumeSource` or just fetch these secrets from `StorageClass.Parameters`?

cc @gnufied

rootfs: resizer needs admin credentials, not publish/stage credentials (used by attacher), in most cases.

pkg/csi/client.go
```diff
@@ -0,0 +1,198 @@
...
+	Expand(ctx context.Context, volumeID string, requestBytes int64, secrets map[string]string) (int64, bool, error)
+
+	// Probe checks that the CSI driver is ready to process requests
+	Probe(ctx context.Context) error
+}
+
+// New creates a new CSI client.
+func New(address string, timeout time.Duration) (Client, error) {
+	conn, err := newGRPCConnection(address, timeout)
+	if err != nil {
```
gnufied: Lets use connection interface from https://github.com/kubernetes-csi/csi-lib-utils/pull/11 ?

pkg/resizer/csi_resizer.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/kubernetes-csi/external-resizer/pkg/csi"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+// NewCSIResizer creates a new resizer responsible for resizing CSI volumes.
+func NewCSIResizer(address string, timeout time.Duration) (Resizer, error) {
```
gnufied: It is bit odd that all these functions take timeout parameter but convert this into a context anyways. Shouldn't we just be passing `Context` here in the first place?

pkg/resizer/csi_resizer.go
gnufied: Lets do what `external-provisioner` does. It also fetches secrets from SC. We could use same secret that are being used for provisioning. cc @msau42

pkg/resizer/csi_resizer.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		if err == wait.ErrWaitTimeout {
+			return errors.New("waiting for driver ready timeout")
+		}
+		return err
+	}
+	return nil
+}
+
+func getDriverName(client csi.Client, timeout time.Duration) (string, error) {
```
gnufied: I would preferably convert these functions to directly accept contexts rather than converting them around.

pkg/resizer/csi_resizer.go
msau42: Which secret needs the be used? Same as provision? Or do we need a new secret for resize?

gnufied: In most cases it will be same as provisioner secret but may be in some cases user have generated secrets which have permission to resize but not to provision? 

looks like `csi.storage.k8s.io/provisioner-secret-name/namespace` is what provisioner is using for fetching secrets. I think we will have to either introduce new parameters to SC or reuse provisioner secrets? Reusing provisioner secrets feel dirty so I guess the only workable long-term solution is to let SC define different secrets for resizing.

msau42: I think having a separate secret for resize makes sense. Should it also be added to the CSI volume spec?

gnufied: you mean `CSIPersistentVolumeSource` ? I am not sure we need resizing secrets there because resizing is not supported for inline volume anyways.

msau42: PV source

gnufied: PV source still refers to `CSIPersistentVolumeSource` right? So no - I do not think we need resizing secrets there.

msau42: It's where we put all the other non provisioning secrets. So you don't need to reference storageclass.

pkg/csi/client.go
mlmhl: OK, these util functions make sense to me, I will update this later.

pkg/resizer/csi_resizer.go
mlmhl: `WithTimeout` returns `WithDeadline(parent, time.Now().Add(timeout))`, so I think we need to create the context just before each CSI operation. If we passed a `Context` to `NewCSIResizer` method, it means the total time consumed by all CSI operations(`waitDriverReady`, `getDriverName`, etc.) should not exceed `timeout`. This is not what we want.

pkg/resizer/csi_resizer.go
mlmhl: OK, I will fetch secrets from SC with a separate name.

gnufied:
@mlmhl can you update this PR with fetching secret from SC and then we can mostly merge this.

pkg/resizer/csi_resizer.go
msau42: Instead of fetching secrets directly from the storageclass, can it be added as a parameter to the CSIPersistentVolumeSource?  That way all the secret fetching/templating logic can be confined to just the external-provisioner.

mlmhl:
@gnufied  I've updated this PR to fetch secrets from `StorageClass`'s parameters, with different name/namespace key, PTAL :)

 ps: We should add these keys to external-provisioner too, as it needs to remove these keys from parameters before passing to the driver, see [this function](https://github.com/kubernetes-csi/external-provisioner/blob/b336e3dbd66c822f32cdf5c6e839f32be2ed7dfc/pkg/controller/controller.go#L641). I will post a PR to do this after this PR merged. cc @msau42 

What's more, I haven't introduce the connection interface in https://github.com/kubernetes-csi/csi-lib-utils due to the dependency conflict: csi-lib-utils requires https://github.com/container-storage-interface/spec `v1.0.0`, but external-resizer requires `master`. I think we can resolve this in a follow up PR once csi-lib-utils are updated to use the latest CSI spec package.

msau42:
I can help update CSI lib utils to use the master CSI spec

mlmhl:
Thanks @msau42 !

gnufied:
@saad-ali should we consider tagging a new CSI release since we need it for resizing?

saad-ali:
@gnufied Yes. Please please start a thread (propose cutting a CSI 1.1 and a timeline) and add it it to the agenda of the next CSI meeting.

gnufied:
lets merge this for now rather than being blocked on csi tag or csi-lib being updated.  Also, for alpha release using secrets from SC is fine, we will update this code to fetch from "new place", once that decision is finalized in https://github.com/kubernetes-csi/external-provisioner/issues/233


/lgtm

Issue #13, opened 2019-02-27
Provider example YAMLs for deploying external-resizer


gnufied:

gnufied:
@mlmhl can you create necessary rbac rules etc that the external controller is going to depend on for deploying external resizer in a pod?

mlmhl:
@gnufied Yeah, of cause! I will submit a PR later.
/assign

PR #16, opened 2019-03-03
add example YAMLs for deploying external-resizer

mlmhl:
ref: https://github.com/kubernetes-csi/external-resizer/issues/13

### DIFF
--- a/deploy/kubernetes/deployment.yaml
+++ b/deploy/kubernetes/deployment.yaml
@@ -1,0 +1,57 @@
+# This YAML file demonstrates how to deploy the external
+# resizer for use with the mock CSI driver. It
+# depends on the RBAC definitions from rbac.yaml.
+
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: csi-resizer
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      external-resizer: mock-driver
+  template:
+    metadata:
+      labels:
+        external-resizer: mock-driver
+    spec:
+      serviceAccount: csi-resizer
+      containers:
+        - name: csi-resizer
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+
+        - name: mock-driver
+          image: quay.io/k8scsi/mock-plugin
+          imagePullPolicy: "IfNotPresent"
+          env:
+            - name: CSI_ENDPOINT
+              value: /var/lib/csi/sockets/pluginproxy/mock.socket
+          volumeMounts:
```
gnufied: Should we be consistent with other external components that mostly use "/csi/csi.sock" ?

mlmhl: Currently only external snapshotter uses `/csi/csi.sock`: https://github.com/kubernetes-csi/external-snapshotter/blob/f59b58d16488bbb9c15a263bf1b2685c3c4c4441/deploy/kubernetes/setup-csi-snapshotter.yaml#L94, both provisioner and attacher use `/var/lib/csi/sockets/pluginproxy/mock.socket`: https://github.com/kubernetes-csi/external-provisioner/blob/f588078fd1f33dd7e9557cceba4c05e3562683fd/deploy/kubernetes/statefulset.yaml#L41, https://github.com/kubernetes-csi/external-attacher/blob/0e6f5f22774f89d2d2c396896c462a4fb3587902/deploy/kubernetes/deployment.yaml#L53.
So I tend to be consistent with the provisioner and attacher as they all use `mock driver`, while snapshotter uses `hostpath`.

PR #23, opened 2019-03-06
relax release tools check

pohly:
The situation with `release-tools` in this repo is a bit special: we have to ignore older commits which added and then removed `release-tools.

/assign @gnufied


pohly:
I can't wait to see whether this now works in Travis CI. Please merge if it does.

Issue #22, opened 2019-03-06
Problem with volumes that support node only expansion


gnufied:
It looks like if a CSI volume supports node only expansion and does not have controller EXPAND_VOLUME capability this resize controller will exit. 

But problem is - in that case the PV will never be updated and currently node side expansion expeects PV to have correct size before calling node side expansion. Also even if Node side expansion handled updating PV, there would be resistance to updating PV from node. 

I am thinking this controller should not exit if CSI volume does not support control-plane volume expansion but should do a noop expansion and update the PV regardless. This is how flexvolume resizing has been implemented.

cc @mlmhl @msau42 @chakri-nelluri

gnufied:
cc @kubernetes-csi/csi-misc

mlmhl:
> I am thinking this controller should not exit if CSI volume does not support control-plane volume expansion but should do a noop expansion and update the PV regardless.

+1

I add a `trivialResizer` in the init implement, and its original intention is to solve this problem. I think we can use a `trivialResizer` instead of `CSIResizer` to handle requests of drivers which do not have controller EXPAND_VOLUME capability. The `trivialResizer` should do nothing but mark the volume as file system resize required.

I will fix this if there is no objection.
/assign

jsafrane:
We're in the same situation as with attacher. Initially, it was solved by `trivialAttacher` in external-attacher that marks all volumes as attached. In reality users did not understand why do they need to run attacher if their driver does not support external attach/detach.

We added a flag into CSIDriver so CSI volume plugin called by A/D controller *knows* that a volume does not support attach/detach and an just return to A/D controller that volume is attached or detacher. No external-attacher is running in this case.

IMO, we need similar flag in CSIDriver for controller-initiated expand (or how do you call it).

mlmhl:
> IMO, we need similar flag in CSIDriver for controller-initiated expand (or how do you call it).

There is a little different between external resizer  and attacher: Even if a driver only supports node expansion, the external resizer still needs to update PVC's conditions and PV's capacity, and kubelet relies on this behavior. cc @gnufied

gnufied:
@jsafrane @mlmhl yeah so if we did add such a flag then perhaps we can update existing in-tree controller to update PV if we know volume does not support resizing.  But lets finalize on this idea in next release when we move to beta. I think for this release having a trivial resizer seems to be fine.

mlmhl:
[...]

PR #26, opened 2019-03-07
Use trivial resizer if only node expansion supported

mlmhl:
If a driver supports node expansion only, use a trivial resizer to just update PV's capacity and PVC's condition, without any meaningful expand operation.

ref: https://github.com/kubernetes-csi/external-resizer/issues/22


pkg/resizer/csi_resizer.go
```diff
@@ -46,47 +41,49 @@ const (
...
-	if !supports {
-		return nil, fmt.Errorf("CSI driver does not support controller resize")
+	if !supportControllerResize {
+		supportsNodeResize, err := supportsNodeResize(csiClient, timeout)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if plugin supports node resize: %v", err)
+		}
+		if supportsNodeResize {
+			klog.Info("The CSI driver supports node resize only, using trivial resizer to handle resize requests")
+			return newTrivialResizer(driverName), nil
```
gnufied: is there a way to get some tests in? There isn't an e2e that tests this as well, so we have this somewhat untested.  We need an e2e or something that tests the scenario.

mlmhl: OK, I will try to add some unit tests first.

mlmhl:
@gnufied Add some unit tests, PTAL :)

gnufied:
thanks. Waiting for tests to be green though..

mlmhl:
Oh sorry, I forgot to check the CI output, the tests are green now.

PR #27, opened 2019-03-12
Use distroless as base image

jsafrane:


jsafrane:
Reason: k8s core components move its base images to distroless/static. See kubernetes/kubernetes#70249 (comment) for details.

mlmhl:
This is LGTM. Let @gnufied make the final decision.

PR #31, opened 2019-03-29
Use leader election library in csi-lib-utils

andrewsykim:
Also updates kubernetes client to v1.14.0 and adds an option to use lease based leader election (highly recommended) over Endpoints. 

/assign @msau42 @xing-yang @jsafrane


cmd/csi-resizer/main.go
```diff
@@ -86,24 +97,42 @@ func main() {
...
-	var leaderElectionConfig *util.LeaderElectionConfig
-	if *enableLeaderElection {
-		if leaderElectionIdentity == nil || *leaderElectionIdentity == "" {
-			klog.Fatal("--leader-election-identity must not be empty")
+	}
+
+	if !*enableLeaderElection {
+		run(context.TODO())
+	} else {
+		if *leaderElectionIdentity == "" {
```
andrewsykim: I would like to get rid of this flag since csi-lib-utils will check the pod hostname if the identity is not explicitly set.

cmd/csi-resizer/main.go
```diff
@@ -42,6 +44,7 @@ var (
 	showVersion = flag.Bool("version", false, "Show version")
 
 	enableLeaderElection      = flag.Bool("leader-election", false, "Enable leader election.")
+	leaderElectionType        = flag.String("leader-election-type", "endpoints", "the type of leader election, options are 'endpoints' (default) or 'lease' (strongly recommended)")
```
msau42: resizer is still alpha (and not officially released yet). Let's just start from lease :-)

andrewsykim: gladly :)

andrewsykim: Going to identity, retry, lease duration & renew flags. Easier to add later than to remove

cmd/csi-resizer/main.go
```diff
@@ -43,36 +43,12 @@ var (
...
-			"before it is replaced by another candidate. This is only applicable if leader "+
-			"election is enabled.")
-	leaderElectionRenewDeadLine = flag.Duration("leader-election-renew-deadline", time.Second*10,
-		"The duration that non-leader candidates will wait after observing a leadership "+
-			"renewal until attempting to acquire leadership of a led but unrenewed leader "+
-			"slot. This is effectively the maximum duration that a leader can be stopped "+
-			"before it is replaced by another candidate. This is only applicable if leader "+
-			"election is enabled.")
+	enableLeaderElection    = flag.Bool("leader-election", false, "Enable leader election.")
+	leaderElectionNamespace = flag.String("leader-election-namespace", "Namespace where this resizer runs.")
```
mlmhl: Did you test this code? The `flag.String` method needs three arguments but you only provides two, this will make the CI test failed.

andrewsykim: good catch, thanks

andrewsykim:
/hold 

Waiting for https://github.com/kubernetes-csi/csi-lib-utils/pull/24

msau42:
Update commit msg and RBAC

deploy/kubernetes/rbac.yaml
```diff
@@ -71,6 +71,9 @@ rules:
 - apiGroups: [""]
```
msau42: Oops can we remove the endpoints rule?

andrewsykim: yes, I'll follow-up

PR #32, opened 2019-04-03
remove permissions to endpoints for example RBAC

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

Endpoints is no longer needed since we are now using Leases for leader election

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -68,9 +68,6 @@ metadata:
   namespace: default
   name: external-resizer-cfg
 rules:
-- apiGroups: [""]
-  resources: ["endpoints"]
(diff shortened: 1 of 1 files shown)


gnufied:
Can you also update rbac policies for external-resizer in in-tree e2e manifests?

andrewsykim:
Sure thing, is that always running on master?

gnufied:
You mean master node? Not necessarily - but then we don't use leader election in e2e, so we should be fine. Code in question btw - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/testing-manifests/storage-csi/external-resizer/rbac.yaml

andrewsykim:
Thanks, opened https://github.com/kubernetes/kubernetes/pull/76091

PR #39, opened 2019-04-25
Implement CSI migration logic for volume resize

### LINKED ISSUE #41: Perform resizing of migrated PVCs

leakingtapan:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This is a continuation of the CSI migration for volume resizing. It implements migration logic for in-tree plugin to use CSI driver for volume resize:
* updated external-resizer to support resizing in-tree volumes types as well as CSI volume. 
* updated external-resizer to handle raw block volume

See https://github.com/kubernetes/community/pull/3624 for WIP migration design for volume resize.

Feature tracking issue: kubernetes/enhancements#625

/cc @ddebroy @davidz627 @gnufied @msau42 

Fixes #41

```release-note
Implement CSI migration logic for volume resizing
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -123,11 +123,27 @@ func (ctrl *resizeController) updatePVC(oldObj, newObj interface{}) {
 	newSize := newPVC.Spec.Resources.Requests[v1.ResourceStorage]
 	oldSize := oldPVC.Spec.Resources.Requests[v1.ResourceStorage]
 
+	newResizerName := newPVC.Annotations[util.VolumeResizerKey]
+	oldResizerName := oldPVC.Annotations[util.VolumeResizerKey]
+
 	// We perform additional checks to avoid double processing of PVCs, as we will also receive Update event when:
 	// 1. Administrator or users may introduce other changes(such as add labels, modify annotations, etc.)
 	//    unrelated to volume resize.
 	// 2. Informer will resync and send Update event periodically without any changes.
-	if newSize.Cmp(oldSize) > 0 {
+	//
+	// We add the PVC into work queue when the new size is larger then the old size
+	// or when the resizer name changes. This is needed for CSI migration for the follow two cases:
+	//
+	// 1. First time a migrated PVC is expanded:
+	// It does not yet have the annotation because annotation is only added by in-tree resizer when it receives a volume
+	// expansion request. So first update event that will be received by external-resizer will be ignored because it won't
+	// know how to support resizing of a "un-annotated" in-tree PVC. When in-tree resizer does add the annotation, a second
+	// update even will be received and we add the pvc to workqueue. If annotation matches the registered driver name in
+	// csi_resizer object, we proceeds with expansion internally or we discard the PVC.
+	// 2. An already expanded in-tree PVC:
+	// An in-tree PVC is resized with in-tree resizer. And later, CSI migration is turned on and resizer name is updated from
+	// in-tree resizer name to CSI driver name.
+	if newSize.Cmp(oldSize) > 0 || newResizerName != oldResizerName {
 		ctrl.addPVC(newObj)
 	}
 }
@@ -249,7 +265,7 @@ func (ctrl *resizeController) pvcNeedResize(pvc *v1.PersistentVolumeClaim) bool 
 
 // pvNeedResize returns true if a pv supports and also requests resize.
 func (ctrl *resizeController) pvNeedResize(pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume) bool {
-	if !ctrl.resizer.CanSupport(pv) {
+	if !ctrl.resizer.CanSupport(pv, pvc) {
 		klog.V(4).Infof("Resizer %q doesn't support PV %q", ctrl.name, pv.Name)
 		return false
 	}
--- a/pkg/resizer/csi_resizer.go
+++ b/pkg/resizer/csi_resizer.go
@@ -23,12 +23,15 @@ import (
 	"time"
 
 	"github.com/kubernetes-csi/external-resizer/pkg/csi"
+	"github.com/kubernetes-csi/external-resizer/pkg/util"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+
+	csitranslationlib "k8s.io/csi-translation-lib"
 	"k8s.io/klog"
 )
 
@@ -107,7 +110,18 @@ func (r *csiResizer) Name() string {
 	return r.name
 }
 
-func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
+// CanSupport returns whether the PV is supported by resizer
+// Resizer will resize the volume if it is CSI volume or is migration enabled in-tree volume
+func (r *csiResizer) CanSupport(pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) bool {
+	resizerName := pvc.Annotations[util.VolumeResizerKey]
+	// resizerName will be CSI driver name when CSI migration is enabled
+	// otherwise, it will be in-tree plugin name
+	// r.name is the CSI driver name, return true only when they match
+	// and the CSI driver is migrated
+	if csitranslationlib.IsMigratedCSIDriverByName(r.name) && resizerName == r.name {
+		return true
+	}
+
 	source := pv.Spec.CSI
 	if source == nil {
 		klog.V(4).Infof("PV %s is not a CSI volume, skip it", pv.Name)
@@ -120,14 +134,32 @@ func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
 	return true
 }
 
+// Resize resizes the persistence volume given request size
+// It supports both CSI volume and migrated in-tree volume
 func (r *csiResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quantity) (resource.Quantity, bool, error) {
 	oldSize := pv.Spec.Capacity[v1.ResourceStorage]
 
-	source := pv.Spec.CSI
-	if source == nil {
-		return oldSize, false, errors.New("not a CSI volume")
+	var volumeID string
+	var source *v1.CSIPersistentVolumeSource
+	if pv.Spec.CSI != nil {
+		// handle CSI volume
+		source = pv.Spec.CSI
+		volumeID = source.VolumeHandle
+	} else {
+		if csitranslationlib.IsMigratedCSIDriverByName(r.name) {
+			// handle migrated in-tree volume
+			csiPV, err := csitranslationlib.TranslateInTreePVToCSI(pv)
+			if err != nil {
+				return oldSize, false, fmt.Errorf("failed to translate persistent volume: %v", err)
+			}
+			source = csiPV.Spec.CSI
+			volumeID = source.VolumeHandle
+		} else {
+			// non-migrated in-tree volume
+			return oldSize, false, fmt.Errorf("volume %v is not migrated to CSI", pv.Name)
+		}
 	}
-	volumeID := source.VolumeHandle
+
 	if len(volumeID) == 0 {
 		return oldSize, false, errors.New("empty volume handle")
 	}
@@ -148,6 +180,7 @@ func (r *csiResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quanti
 	if err != nil {
 		return oldSize, nodeResizeRequired, err
 	}
+
 	return *resource.NewQuantity(newSizeBytes, resource.BinarySI), nodeResizeRequired, err
 }
 
--- a/pkg/resizer/csi_resizer_test.go
+++ b/pkg/resizer/csi_resizer_test.go
@@ -1,14 +1,16 @@
 package resizer
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 	"time"
 
 	"github.com/kubernetes-csi/external-resizer/pkg/csi"
+	"github.com/kubernetes-csi/external-resizer/pkg/util"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
@@ -56,7 +58,7 @@ func TestNewResizer(t *testing.T) {
 			Error: resizeNotSupportErr,
 		},
 	} {
-		client := csi.NewMockClient(c.SupportsNodeResize, c.SupportsControllerResize, c.SupportsPluginControllerService)
+		client := csi.NewMockClient("mock", c.SupportsNodeResize, c.SupportsControllerResize, c.SupportsPluginControllerService)
 		k8sClient, informerFactory := fakeK8s()
 		resizer, err := NewResizerFromClient(client, 0, k8sClient, informerFactory)
 		if err != c.Error {
@@ -89,7 +91,7 @@ func TestResizeWithSecret(t *testing.T) {
 		},
 	}
 	for _, tc := range tests {
-		client := csi.NewMockClient(true, true, true)
+		client := csi.NewMockClient("mock", true, true, true)
 		secret := makeSecret("some-secret", "secret-namespace")
 		k8sClient := fake.NewSimpleClientset(secret)
 		pv := makeTestPV("test-csi", 2, "ebs-csi", "vol-abcde", tc.hasExpansionSecret)
@@ -117,7 +119,7 @@ func TestResizeWithSecret(t *testing.T) {
 
 func makeSecret(name string, namespace string) *v1.Secret {
 	return &v1.Secret{
-		ObjectMeta: meta.ObjectMeta{
+		ObjectMeta: metav1.ObjectMeta{
 			Name:            name,
 			Namespace:       namespace,
 			UID:             "23456",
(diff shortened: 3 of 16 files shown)


pkg/resizer/csi_resizer.go
```diff
@@ -118,7 +120,13 @@ func (r *csiResizer) Name() string {
 	return r.name
 }
 
+// CanSupport returns whether the PV is supported by resizer
+// Resizer will resize the volume if it is CSI volume or is migration enabled in-tree volume
 func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
+	if csitranslationlib.IsMigratedCSIDriverByName(r.name) /* && migration enabled*/ {
```
gnufied: shouldn't this still verify if volume being resized matches registered driver name?

pkg/resizer/csi_resizer.go
```diff
@@ -131,14 +139,30 @@ func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
...
+			// handle migrated in-tree volume
+			csiPV, err := csitranslationlib.TranslateInTreePVToCSI(pv)
+			if err != nil {
+				return oldSize, false, fmt.Errorf("failed to translate persistenc volume: %v", err)
+			}
+			volumeID = csiPV.Spec.CSI.VolumeHandle
+		} else {
+			// non-migrated in-tree volume
+			return oldSize, false, errors.New("un-supported volume type")
+		}
```
gnufied: Can we add some tests for this please?

davidz627:
@leakingtapan is there anything I can do to help unblock this PR?

pkg/resizer/csi_resizer.go
```diff
@@ -131,14 +139,30 @@ func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
...
+		if csitranslationlib.IsMigratedCSIDriverByName(r.name) /* && migration enabled*/ {
+			// handle migrated in-tree volume
+			csiPV, err := csitranslationlib.TranslateInTreePVToCSI(pv)
+			if err != nil {
+				return oldSize, false, fmt.Errorf("failed to translate persistenc volume: %v", err)
+			}
+			volumeID = csiPV.Spec.CSI.VolumeHandle
+		} else {
+			// non-migrated in-tree volume
+			return oldSize, false, errors.New("un-supported volume type")
```
davidz627: would be nice to have more details in this error, like what type was unsupported (or at least that it is a non-migrated in-tree volume type)

leakingtapan:
Finished the rest of the logic and implemented unit test. PTAL

pkg/resizer/csi_resizer.go
```diff
@@ -33,12 +33,19 @@ import (
...
+
+	csitranslationlib "k8s.io/csi-translation-lib"
 	"k8s.io/klog"
 )
 
 const (
 	resizerSecretNameKey      = "csi.storage.k8s.io/resizer-secret-name"
 	resizerSecretNamespaceKey = "csi.storage.k8s.io/resizer-secret-namespace"
+
+	// VolumeResizerKey is key that will be used to store resizer used
```
davidz627: by "store resizer used" do you mean the value associated with the key will be the *resizer name* or some other field that represents the resizer. Could you clarify the comment a bit.

davidz627: also would be useful to note what populates this annotation and whether we can always expect it to be there or what to do if it doesn't exist

pkg/resizer/csi_resizer_test.go
```diff
@@ -66,8 +71,218 @@ func TestNewResizer(t *testing.T) {
...
+		})
+	}
+}
+
+func quantityGB(i int) resource.Quantity {
+	q := resource.NewQuantity(int64(i*1024*1024), resource.BinarySI)
+	return *q
+}
+
+func createPVC(requestGB, capacityGB int, resizerName string) *v1.PersistentVolumeClaim {
```
davidz627: seems like `requestGB` and `capacityGB` are not necessary inputs to this function. We don't actually test different values for these so we could just make them consts. Ditto throughout for the capacity numbers we're passing in. Everything could just be one flat number.

If you *must* pass them in, lets use consts at the caller level so it's easy to tell what each number means

pkg/resizer/csi_resizer_test.go
```diff
@@ -66,8 +71,218 @@ func TestNewResizer(t *testing.T) {
...
+			driverName: "ebs.csi.aws.com",
+			pv:         createEBSPV(1),
+			pvc:        createPVC(2, 1, "ebs.csi.aws.com"),
+			canSupport: true,
+		},
+		{
+			name:       "EBS PV/PVC is not supported when migartion is disabled",
+			driverName: "ebs.csi.aws.com",
+			pv:         createEBSPV(1),
+			pvc:        createPVC(2, 1, "kubernetes.io/aws-ebs"),
```
davidz627: its not clear to me why setting the pvc annotation to this means "migration is disabled", this could probably be fixed by more comments on the annotation key that I mentioned in the other comment

pkg/resizer/csi_resizer.go
```diff
@@ -131,14 +145,30 @@ func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
...
-	if source == nil {
-		return oldSize, false, errors.New("not a CSI volume")
+	var volumeID string
+	if pv.Spec.CSI != nil {
+		// handle CSI volume
+		source := pv.Spec.CSI
+		volumeID = source.VolumeHandle
+	} else {
+		if csitranslationlib.IsMigratedCSIDriverByName(r.name) {
+			// handle migrated in-tree volume
```
ddebroy: Can this logic from L154 - L170 be compacted in the following manner (do the translation first if necessary and then simply grab the volumeID from the translated pv.Spec.CSI):

```
if csitranslationlib.IsMigratedCSIDriverByName(r.name) {
	pv, err = csitranslationlib.TranslateInTreePVToCSI(pv); err != nil {
		return oldSize, false, fmt.Errorf("failed to translate persistent volume: %v", err)
	}
}
source := pv.Spec.CSI
if source == nil {
	return oldSize, false, errors.New("cannot process a volume without a CSI source")
}
volumeID := source.VolumeHandle

```

pkg/resizer/csi_resizer.go
gnufied: FWIW it is populated by in-tree volume resizer and if in-tree migration is enabled for the plugin, the value corresponding to this key should be csi driver name (even though PV spec is in-tree).

gnufied: Also if migration is not enabled, the value of this key should be in-tree k8s driver.

leakingtapan: udpated

pkg/resizer/csi_resizer_test.go
leakingtapan: yeah. It means, when CSI driver is installed, but migration is not enabled. Updated the comment for the key

pkg/resizer/csi_resizer.go
leakingtapan: Good idea. Updated

leakingtapan: hmmm. Just tested and find out this won't work for the case of CSI volume since `csitranslationlib.IsMigratedCSIDriverByName(r.name)` returns true and it will attempt to translate a CSI volume to CSI

leakingtapan:
After testing the new change with synchronization, I found a design flaw of using resizer key from annotation. When the new resizer key `volume.kubernetes.io/storage-resizer` is used, since both in-tree resizer and external resizer is trying to resize the volume at the same time, the external-resizer won't be able to read the resizer key since it's not even written by in-tree resizer yet. 

The current workaround is fallback to provisioner key. WDYT? 

/cc @davidz627 @ddebroy @gnufied

leakingtapan:
I also added support for resizing raw block volume, and it works with current change

gnufied:
Do we know if provisioner key annotation is added to all existing PVCs too? I quickly looked at the code and it looks like it is added only for new PVCs..

gnufied:
I am thinking we can fix this problem by looking at `update` event and if an update event causes resizer annotation to be added (by comparing old vs new pvc), then we can add the PVC to workqueue. If PVC annotation happens to not match CSI driver, it will be removed from workqueue and PVC will be processed, if annotation matches CSI driver name then we will proceed with volume expansion.

@leakingtapan wdyt?

leakingtapan:
@gnufied addressed comment and updated the condition is add PVC into work queue when the reszier annotation is added from `update` event

pkg/resizer/csi_resizer_test.go
```diff
@@ -160,8 +314,68 @@ func makeTestPV(name string, sizeGig int, driverName, volID string, withSecret b
 	return pv
 }
 
+func createPV(capacityGB int) *v1.PersistentVolume {
```
gnufied: minor nit - can we reuse function `makeTestPV` instead of defining new one?

pkg/resizer/csi_resizer_test.go
```diff
@@ -160,8 +314,68 @@ func makeTestPV(name string, sizeGig int, driverName, volID string, withSecret b
...
+
+}
+
 func fakeK8s() (kubernetes.Interface, informers.SharedInformerFactory) {
 	client := fake.NewSimpleClientset()
 	informerFactory := informers.NewSharedInformerFactory(client, 0)
 	return client, informerFactory
 }
+
+func createEBSPV(capacityGB int) *v1.PersistentVolume {
```
gnufied: lets call these functions `createInTreeEBSPV` for clear intention.

pkg/resizer/csi_resizer_test.go
```diff
@@ -160,8 +314,68 @@ func makeTestPV(name string, sizeGig int, driverName, volID string, withSecret b
...
+			PersistentVolumeSource: v1.PersistentVolumeSource{
+				AWSElasticBlockStore: &v1.AWSElasticBlockStoreVolumeSource{
+					VolumeID: "testVolumeId",
+				},
+			},
+		},
+	}
+}
+
+func createGCEPDPV(capacityGB int) *v1.PersistentVolume {
```
gnufied: same

leakingtapan:
Updated PR

pkg/controller/controller.go
```diff
@@ -123,11 +123,18 @@ func (ctrl *resizeController) updatePVC(oldObj, newObj interface{}) {
...
+
 	// We perform additional checks to avoid double processing of PVCs, as we will also receive Update event when:
 	// 1. Administrator or users may introduce other changes(such as add labels, modify annotations, etc.)
 	//    unrelated to volume resize.
 	// 2. Informer will resync and send Update event periodically without any changes.
-	if newSize.Cmp(oldSize) > 0 {
+	//
+	// We add the PVC into work queue when the new size is larger then the old size
+	// or when the resizer name annotation is missing from old PVC
+	// but present in the new PVC. This is needed for CSI migration
```
msau42: Can you describe more in detail why it is needed for migration?  Even when migration is enabled, wouldn't the new and old sizes be different?

pkg/resizer/csi_resizer.go
```diff
@@ -120,14 +130,32 @@ func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
...
+			csiPV, err := csitranslationlib.TranslateInTreePVToCSI(pv)
+			if err != nil {
+				return oldSize, false, fmt.Errorf("failed to translate persistent volume: %v", err)
+			}
+			source = csiPV.Spec.CSI
+			volumeID = source.VolumeHandle
+		} else {
+			// non-migrated in-tree volume
+			return oldSize, false, fmt.Errorf("volume %v is not migrated to CSI", pv.Name)
```
msau42: How do we get here? The annotation is set with the name that the csi plugin is configured for but the migration library doesn't support it?

pkg/resizer/csi_resizer.go
```diff
@@ -148,7 +176,12 @@ func (r *csiResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quanti
 	if err != nil {
 		return oldSize, nodeResizeRequired, err
 	}
-	return *resource.NewQuantity(newSizeBytes, resource.BinarySI), nodeResizeRequired, err
+
+	if pv.Spec.VolumeMode == nil || *pv.Spec.VolumeMode == v1.PersistentVolumeFilesystem {
+		return *resource.NewQuantity(newSizeBytes, resource.BinarySI), nodeResizeRequired, err
+	}
+
+	return *resource.NewQuantity(newSizeBytes, resource.BinarySI), false, err
```
msau42: Are there any block plugins that only implement node resize?

pkg/resizer/csi_resizer.go
```diff
@@ -107,7 +110,14 @@ func (r *csiResizer) Name() string {
 	return r.name
 }
 
-func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
+// CanSupport returns whether the PV is supported by resizer
+// Resizer will resize the volume if it is CSI volume or is migration enabled in-tree volume
+func (r *csiResizer) CanSupport(pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) bool {
+	resizerName := pvc.Annotations[util.VolumeResizerKey]
+	if csitranslationlib.IsMigratedCSIDriverByName(r.name) && resizerName == r.name {
```
msau42: Is the annotation the in-tree name or csi name?

Is r.name the in-tree name or the csi name?

msau42: (can you add a comment)

pkg/controller/controller.go
gnufied: the problem is - when first time a migrated PVC is expanded, it does not yet have the annotation because annotation is only added by in-tree resizer when it receives a volume expansion request. So first update event that will be received by external-resizer will be ignored because it won't know how to support resizing of a "un-annotated" in-tree PVC. 

When in-tree resizer does add the annotation, a second update even will be received and we add the pvc to workqueue. If annotation matches the registered driver name in csi_resizer object, we proceeds with expansion internally or we discard the PVC.

msau42: To determine if we should process the PVC, should we be comparing spec/status of size, instead of new/old PVC size?  How do we retry resizing operation if it fails?

Regardless, if this is the way it needs to be, let's explain it in a comment.

leakingtapan: Thx @gnufied for detailed explanation, added you reply into the comment.

pkg/resizer/csi_resizer.go
leakingtapan: This will be an in-tree PV that doesn't support CSI migration yet. This should be already gated by `CanSupport` check earlier, but still adding the else branch for the sake of in case this happens unintentionally, or because of some bug

pkg/resizer/csi_resizer.go
leakingtapan: Not that Im aware of, could it be a case for local volume plugin? And how is this going to affect this highlighted line?

pkg/resizer/csi_resizer.go
leakingtapan: added more comments

leakingtapan:
Added more comments @msau42

pkg/controller/controller.go
gnufied: We do compare spec/status size later - https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/controller.go#L213 but having new vs old size check in update prevents processing of events which aren't size related updates for PVC.

pkg/resizer/csi_resizer.go
msau42: There is nothing in the csi spec that limits node expansion to only filesystem volumes.  For a block plugin that only supports node resize, then is setting nodeResizeRequired=false going to cause nodeResize to never be called?

Should we instead always call node expansion (based off of nodeResizeRequired)?  If the plugin gets a node expand call for a block device, can they just return success immediately?

pkg/resizer/csi_resizer.go
```diff
@@ -107,7 +110,18 @@ func (r *csiResizer) Name() string {
 	return r.name
 }
 
-func (r *csiResizer) CanSupport(pv *v1.PersistentVolume) bool {
+// CanSupport returns whether the PV is supported by resizer
+// Resizer will resize the volume if it is CSI volume or is migration enabled in-tree volume
+func (r *csiResizer) CanSupport(pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) bool {
+	resizerName := pvc.Annotations[util.VolumeResizerKey]
+	// resizerName will be CSI driver name when CSI migration is enabled
```
msau42: This is more of a question for in-tree, but does the annotation get updated if we turn on migration after the PVC has already been annotated with the in-tree name?

gnufied: It does resets the annotation if migration has been enabled after annotation was already set.

pkg/resizer/csi_resizer.go
gnufied: I agree that the responsibility of determining whether node expansion is needed or not should be delegated to the plugin, but in control-plane the plugin may not have the necessary information to know that volume is being used in block device mode. 

For now - I think it makes sense to return whatever `nodeResizeRequired` value was returned by the plugin. In future I am thinking we may want to update CSI spec to pass volume capability with `ControllerExpandVolume` request.

pkg/controller/controller.go
```diff
@@ -123,11 +123,23 @@ func (ctrl *resizeController) updatePVC(oldObj, newObj interface{}) {
...
+	//
+	// We add the PVC into work queue when the new size is larger then the old size
+	// or when the resizer name annotation is missing from old PVC but present in the new PVC.
+	// This is needed for CSI migration since the first time a migrated PVC is expanded,
+	// it does not yet have the annotation because annotation is only added by in-tree resizer when it receives a volume
+	// expansion request. So first update event that will be received by external-resizer will be ignored because it won't
+	// know how to support resizing of a "un-annotated" in-tree PVC. When in-tree resizer does add the annotation, a second
+	// update even will be received and we add the pvc to workqueue. If annotation matches the registered driver name in
+	// csi_resizer object, we proceeds with expansion internally or we discard the PVC.
+	if newSize.Cmp(oldSize) > 0 || (newResizerName != "" && oldResizerName == "") {
```
gnufied: Given https://github.com/kubernetes-csi/external-resizer/pull/39#discussion_r296868667 I think it makes sense to check if `newResizerName != oldResizeName`

pkg/resizer/csi_resizer.go
msau42: I think the question here is more related to node expand.  It looks like we don't pass volume capability to node expansion, and maybe we need to add that?

I guess what a plugin could do now is check if the volume has a filesystem on it. If not, then assume it's raw block, otherwise assume it's filesystem.

pkg/resizer/csi_resizer.go
leakingtapan: Good question! @gnufied do you know the answer to this?

pkg/resizer/csi_resizer.go
gnufied: Shouldn't a plugin be able to determine that given `target_path` in `NodeExpandVolume` RPC call is a device or a mounted filesystem?

davidz627: agree with @gnufied here, `ControllerExpand` should have volume capability in the future so the driver can determine `nodeResizeRequired` better.

`NodeExpandVolume` may be determinable by the `target_path` but it might be nice to know the volume capability too.

Either way, for this PR I think just returning the plugins `nodeResizeRequired` makes sense instead of overwriting with `false`.

pkg/controller/controller.go
leakingtapan: Updated

pkg/resizer/csi_resizer.go
msau42: Unsure if we should be dynamically determining whether or not to call node expand. It is simpler to base it strictly on the plugin's capabilities. Especially across controller and node services, this makes testing/supporting version skew more complicated.

PR #40, opened 2019-04-26
Add controller tests

### LINKED ISSUE #5: Add unit test for external resize controller
We need some basic unit tests and sanity checks for controller.

jarrpa:
**What type of PR is this?**
/kind bug

Introduces unit tests for the resizer controller.

Fixes #5 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -1,0 +1,174 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/kubernetes-csi/external-resizer/pkg/csi"
+	"github.com/kubernetes-csi/external-resizer/pkg/resizer"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestController(t *testing.T) {
+	for _, test := range []struct {
+		Name string
+		PVC  *v1.PersistentVolumeClaim
+		PV   *v1.PersistentVolume
+
+		CreateObjects bool
+		NodeResize    bool
+	}{
+		{
+			Name: "Invalid key",
+			PVC:  invalidPVC(),
+		},
+		{
+			Name: "PVC not found",
+			PVC:  createPVC(1, 1),
+		},
+		{
+			Name:          "PVC doesn't need resize",
+			PVC:           createPVC(1, 1),
+			CreateObjects: true,
+		},
+		{
+			Name:          "PV not found",
+			PVC:           createPVC(2, 1),
+			CreateObjects: true,
+		},
+		{
+			Name:          "PV doesn't need resize",
+			PVC:           createPVC(2, 1),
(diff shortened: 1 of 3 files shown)


jarrpa:
@gnufied Test cleared. Could this get approval?

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return pvc
+}
+
+func quantityGB(i int) resource.Quantity {
+	q := resource.NewQuantity(int64(i*1024*1024), resource.BinarySI)
+	return *q
+}
+
+func pvc(requestGB, capacityGB int) *v1.PersistentVolumeClaim {
```
gnufied: can we name this `getPVC` or something?

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Status: v1.PersistentVolumeClaimStatus{
+			Phase: v1.ClaimBound,
+			Capacity: map[v1.ResourceName]resource.Quantity{
+				v1.ResourceStorage: capacity,
+			},
+		},
+	}
+}
+
+func pv(capacityGB int) *v1.PersistentVolume {
```
gnufied: ditto

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return &v1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "testPV",
+		},
+		Spec: v1.PersistentVolumeSpec{
+			Capacity: map[v1.ResourceName]resource.Quantity{
+				v1.ResourceStorage: capacity,
+			},
+		},
```
gnufied: Can we get this PV to have `CSIPersistentVolumeSource` field, so as it is supported by the resizer? As it is - mock driver's `Expand` call is never made because PVs don't have CSI field.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+func TestController(t *testing.T) {
+	for _, test := range []struct {
+		Name string
+		PVC  *v1.PersistentVolumeClaim
+		PV   *v1.PersistentVolume
+
+		CreateObjects bool
+		NodeResize    bool
```
gnufied: should we add a field here for explicitly testing scenarios where we expect failure? As it is - we only test scenarios which aren't supposed to fail.

jarrpa: All the failure scenarios I saw were either guarding against underlying kube problems (which I don't know how to mock) or resizer problems (which I believe should be handled in the resizer unit tests). Any advice on how of if either of those two categories should be tested would be appreciated.

pkg/controller/controller_test.go
jarrpa: Done.

pkg/controller/controller_test.go
jarrpa: Done.

pkg/controller/controller_test.go
jarrpa: Done.

pkg/controller/controller_test.go
gnufied: iirc - mock client supports adding reactors to certain actions on kube client. We could use that to test more custom stuff. For example, I was wondering if for test case where PV should be updated, if we can actually test if PV got updated. That would be awesome.

gnufied:
lets merge this for now. More tests are better than less tests. I will fix up controller tests to actually check for pv updates in a follow up PR.

/lgtm

PR #47, opened 2019-08-13
Mark PVC for FS resize iff it's not Block

### LINKED ISSUE #46: Block volumes are marked with FileSystemResizeRequired
The in-tree controller, after finishing Controller expand, checks not only if the plugin reports that FS resize is required but also if the volume is mode FS or not: https://github.com/kubernetes/kubernetes/blob/1dac5fd14a54ac4972339dbe55f9f03688fd7542/pkg/volume/util/operationexecutor/operation_generator.go#L1578 before it marks volumes FileSystemResizeRequired. 

The external-resizer should do the same.

This was caught by the e2e test
`[Driver: aws] [Testpattern: Dynamic PV (block volmode)(allowExpansion)] volume-expand [It] Verify if offline PVC expansion works`

/assign

wongma7:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Don't mark Block PVs as needing FS resize

Fixes #46

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -327,7 +327,7 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 			return err
 		}
 
-		if fsResizeRequired {
+		if fsResizeRequired && !(pvc.Spec.VolumeMode != nil && *pvc.Spec.VolumeMode == v1.PersistentVolumeBlock) {
 			// Resize volume succeeded and need to resize file system by kubelet, mark it as file system resizing required.
 			return ctrl.markPVCAsFSResizeRequired(pvc)
 		}


pkg/controller/controller.go
```diff
@@ -327,7 +327,7 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 			return err
 		}
 
-		if fsResizeRequired {
+		if fsResizeRequired && !(pvc.Spec.VolumeMode != nil && *pvc.Spec.VolumeMode == v1.PersistentVolumeBlock) {
```
gnufied: We explicitly made a choice to not skip `NodeExpandVolume` if volume type is of block type. The reasoning is - the spec does not stipulate that `NodeExpandVolume` will not be called if volume is of block type. Only a plugin can decide whether `NodeExpandVolume` can be skipped or not.

This is pending on a fix in kubelet - https://github.com/kubernetes/kubernetes/issues/79990 too.

wongma7: Okay, I still think there is a fix to be made here, to not put FileSystemResizePending & FileSystemResizeRequired on block volumes. NodeExpand can still be called but the user-facing things should be accurate

gnufied: Yeah I think the naming of conditions is bit wrong and pre-dates CSI spec . I will look into renaming those conditions to be more accurate.

wongma7: Hm, how on the plugin side can I decide whether to skip NodeExpandVolume? When it's called, I only get volume ID and path, it doesn't tell me if the CO intends for the volume to be used as block or not. If it's block, I'd like to no-op.

wongma7: ControllerExpandVolume won't know if the volume is block or not either, so it can't return false for node expansion required

msau42: I thought pd driver skipped NodeExpand based on volume capability but I was wrong: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/372

cc @davidz627

gnufied: I think I am in favour of adding supplying volume capabilities into both `ControllerExpandVolume` and `NodeExpandVolume`. `NodeExpandVolume` has a clear use case where we might not want to fs resize the volume even if it has file system on it - if volume is being used in block mode.

leakingtapan: This is a similar change to the one I made during resize migration: https://github.com/kubernetes-csi/external-resizer/pull/39#discussion_r296879475

If we want SP to decide if to do `NodeExpandVolume`, not sure if we still need this change here

leakingtapan:
is this still needed? @wongma7

wongma7:
I don't remember exactly, but I think https://github.com/kubernetes/kubernetes/pull/81429 fixed it. So yeah I'll close.

PR #53, opened 2019-09-13
Update the image and enable e2e runs

gnufied:
/kind cleanup

bumping hostpath driver version to check expansions tests

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,5 +1,6 @@
 #! /bin/bash
 
+CSI_PROW_KUBERNETES_VERSION=latest
+CSI_PROW_E2E_VERSION=master
 . release-tools/prow.sh
-
 main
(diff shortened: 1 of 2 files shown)


gnufied:
cc @pohly

gnufied:
Looks like overriding e2e version seems to have worked but as dicussed they are being skipped now. I will enable the tests.

.prow.sh
```diff
@@ -1,5 +1,6 @@
 #! /bin/bash
 
+CSI_PROW_KUBERNETES_VERSION=latest
```
msau42: Can you add a TODO, or open an issue to remove this once we add 1.16 jobs?

gnufied: opened

pohly: See https://github.com/kubernetes-csi/external-resizer/issues/54

PR #58, opened 2019-10-10
Add changelog for 0.3.0

gnufied:
Add changelog for 0.3.0

cc @msau42


CHANGELOG-0.3.0.md
```diff
@@ -0,0 +1,18 @@
+# v0.3.0
```
msau42: Can you actually make the filename based on minor versions, not patch verisons? So CHANGELOG-0.3.md

gnufied: Do we want rename existing changelog files too?

msau42: yeah that sounds good

PR #61, opened 2019-10-25
Verify claimref associated with PVs before resizing

gnufied:
Verify claimref associated with PV before allowing volume expansion.

```release-note
Verify claimref associated with PV before allowing volume expansion.
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,0 +1,4 @@
+#! /bin/bash
+
+. release-tools/prow.sh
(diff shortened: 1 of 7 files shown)


gnufied:
Because of the way, prow jobs have been setup for CSI sidecars, it is going to be hard to have a green build on this branch. I am going to close this PR, because our current recommendation for csi-resizer is to use latest released version.

PR #65, opened 2019-12-09
Omit extranous dash with CSI timeout parameter

timoreimann:
**What type of PR is this?**
/kind documentation

```release-note
NONE
```

/sig-storage


timoreimann:
CI failure is:

```
fail("Go toolchain does not support c...")
```

So probably not due to my change, yet wondering if a flake that persist retrying would overcome eventually. 🤔

msau42:
Our CI Is broken: #66 

I will force merge for now because this is purely a readme change and doesn't impact the tests

PR #70, opened 2020-01-08
Add volume capability to ControllerExpandVolume CSI call

gnufied:
xref https://github.com/kubernetes/kubernetes/issues/86923

Bumps CSI version and adds new fields to ControllerExpandVolume request

```release-note
Pass volume capability to ControllerExpandVolume RPC call
```

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -188,6 +188,7 @@ func createPV(capacityGB int, pvcName, pvcNamespace string, pvcUID types.UID) *v
 			Name: "testPV",
 		},
 		Spec: v1.PersistentVolumeSpec{
+			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
 			Capacity: map[v1.ResourceName]resource.Quantity{
 				v1.ResourceStorage: capacity,
 			},
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -28,10 +28,10 @@ func TestController(t *testing.T) {
 		PVC  *v1.PersistentVolumeClaim
 		PV   *v1.PersistentVolume
 
-		CreateObjects bool
-		NodeResize    bool
-		CallCSIExpand bool
-		blockVolume   bool
+		CreateObjects     bool
+		NodeResize        bool
+		CallCSIExpand     bool
+		expectBlockVolume bool
 	}{
 		{
 			Name:          "Invalid key",
@@ -89,13 +89,13 @@ func TestController(t *testing.T) {
 			CallCSIExpand: true,
 		},
 		{
-			Name:          "Block Resize PVC with FS resize",
-			PVC:           createPVC(2, 1),
-			PV:            createPV(1, "testPVC", "test", "foobar", &blockVolumeMode),
-			CreateObjects: true,
-			NodeResize:    true,
-			CallCSIExpand: true,
-			blockVolume:   true,
+			Name:              "Block Resize PVC with FS resize",
+			PVC:               createPVC(2, 1),
+			PV:                createPV(1, "testPVC", "test", "foobar", &blockVolumeMode),
+			CreateObjects:     true,
+			NodeResize:        true,
+			CallCSIExpand:     true,
+			expectBlockVolume: true,
 		},
 	} {
 		client := csi.NewMockClient("mock", test.NodeResize, true, true)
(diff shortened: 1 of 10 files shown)


pkg/resizer/csi_resizer.go
```diff
@@ -190,16 +198,82 @@ func (r *csiResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quanti
...
+	newSizeBytes, nodeResizeRequired, err := r.client.Expand(ctx, volumeID, requestSize.Value(), secrets, capability)
 	if err != nil {
 		return oldSize, nodeResizeRequired, err
 	}
 
 	return *resource.NewQuantity(newSizeBytes, resource.BinarySI), nodeResizeRequired, err
 }
 
+// GetVolumeCapabilities returns volumecapability from PV spec
+func GetVolumeCapabilities(pvSpec v1.PersistentVolumeSpec) (*csilib.VolumeCapability, error) {
```
davidz627: unit test for this?

pkg/controller/controller_test.go
```diff
@@ -137,6 +149,19 @@ func TestController(t *testing.T) {
 		if !test.CallCSIExpand && expandCallCount > 0 {
 			t.Fatalf("for %s: expected no csi expand call, received csi expansion request", test.Name)
 		}
+
+		if test.CallCSIExpand {
+			usedCapability := client.GetCapability()
+			if test.blockVolume {
```
davidz627: nit: `expectBlockVolume`

Also these conditionals would be easier to read if stuck together `if test.expectBlockVolum && usedCapability.GetBlock() == nil`

davidz627:
2 small comments, lgtm besides

pkg/controller/controller_test.go
gnufied: fixed

pkg/resizer/csi_resizer.go
gnufied: fixed

gnufied:
@msau42 fixed the PR. PTAL.

pkg/resizer/csi_resizer.go
```diff
@@ -36,6 +37,10 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	defaultFSType = "ext4"
```
xing-yang: Given what we have discussed here: https://github.com/kubernetes-csi/external-provisioner/pull/400/files#diff-38b23914da9991d96c9ec4fc3c2525aaR76, should this also be configurable and allow default to be "none"?
@msau42 ^^

gnufied: For control-plane volume expansion it should *not* ideally matter. This exists mostly for being on same page as external-attacher , what we really want to pass is block vs fs type.

gnufied:
putting on hold so as @xing-yang comment gets addressed.

/hold

pkg/resizer/csi_resizer.go
gnufied: Let me know if this is okay.  Feel free to remove the hold if you are fine with this.

xing-yang: I think resizer should not have a defaultFSType.  Just keep "fsType := pvSpec.CSI.FSType" but don't set a default to be consistent with the proposed change in external-provisioner.  Otherwise we could potentially have a default fstype set to ext4 while driver internally use a different type.

In the external-provisioner, if we let driver set default to "none", driver will determine internally what fstype to use. 
Unfortunately this fstype is not returned by CreateVolumeResponse in CSI.  So pvSpec.CSI.FSType could be not set even if driver has used a non-ext4 type internally.  If we change pvSpec.CSI.FSType to ext4 in resizer in this case, that will be the wrong fstype.

gnufied: okay I have dropped default fstype from this PR.

pkg/resizer/csi_resizer.go
```diff
@@ -190,16 +194,79 @@ func (r *csiResizer) Resize(pv *v1.PersistentVolume, requestSize resource.Quanti
...
 
+	capability, err := GetVolumeCapabilities(pvSpec)
+	if err != nil {
+		return oldSize, false, fmt.Errorf("failed to get capabilities of volume %s with %v", pv.Name, err)
+	}
+
 	ctx, cancel := timeoutCtx(r.timeout)
 	defer cancel()
-	newSizeBytes, nodeResizeRequired, err := r.client.Expand(ctx, volumeID, requestSize.Value(), secrets)
+	newSizeBytes, nodeResizeRequired, err := r.client.Expand(ctx, volumeID, requestSize.Value(), secrets, capability)
```
msau42, line 204: What happens if you call this against a driver that doesn't have this new field?

gnufied: If the server is compiled (i.e the driver) is compiled with a version of CSI which does not have this field then the driver will not simply see it.

PR #71, opened 2020-01-17
fix incorrect link

windayski:
fix incorrect link

### DIFF
--- a/release-tools/README.md
+++ b/release-tools/README.md
@@ -81,7 +81,7 @@ on what is enabled in Prow, see
 https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-csi
 
 Test results for periodic jobs are visible in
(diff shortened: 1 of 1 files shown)


release-tools/README.md
```diff
@@ -81,7 +81,7 @@ on what is enabled in Prow, see
 https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-csi
 
 Test results for periodic jobs are visible in
-https://testgrid.k8s.io/sig-storage-csi
+https://testgrid.k8s.io/sig-storage-csi-external-resizer
```
msau42: Unfortunately the way the directory structure works is that contents of release-tools are shared across all repos. So we can't have repo-specific links in here.

https://testgrid.k8s.io/sig-storage-csi-ci  should probably be the correct link. And you'll need to make a PR in the csi-release-tools repo instead of here.

Issue #72, opened 2020-02-05
Arg "--csiTimeout" is not consistent with other sidecars


Madhu-1:
other sidecars are using `timeout` as argument example csi-provisioner https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L64

Madhu-1:
CC @gnufied @msau42

Madhu-1:
@jsafrane we may have to consider the your comment https://github.com/kubernetes-csi/external-provisioner/pull/402#discussion_r375213182 here also based on  the release plan

jsafrane:
Yes, it's the same problem. If there is an user that has yaml file (helm chart, operator, ...) for a CSI driver, this yaml file should work after a minor update of any CSI sidecar (1.5 -> 1.6). We can add new options and mark the old ones as deprecated, but the old ones must still work.

jsafrane:
Btw, thanks for keeping the cmdline options consistent!

msau42:
This is fixed by https://github.com/kubernetes-csi/external-resizer/pull/82
/close

PR #82, opened 2020-05-21
Replace `csiTimeout` to `timeout` param for resizer

humblec:
At present the resizer has CSI call timeout value
taken as arg in form of `--csiTimeout` , however all
other sidecars have this parameter as `--timeout`.
This patch replaces `csiTimeout` to `timeout` value

NOTE: This is a breaking change, so we need a new release

```release-note
The `--csiTimeout` flag of external-resizer is obsoleted with this release. Instead `--timeout` flag has been introduced for the same purpose. The   timeout value  accommodates majority of `ControllerExpandVolume` calls. `timeout` value defaults to 10 seconds. From now on, the deployment  templates or  artifacts has to make use of `--timeout` flag instead of obsoleted `--csiTimeout`  flag.
```
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

### DIFF
--- a/cmd/csi-resizer/main.go
+++ b/cmd/csi-resizer/main.go
@@ -20,11 +20,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
+	"time"
+
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"os"
-	"time"
 
 	"k8s.io/client-go/util/workqueue"
 
@@ -44,8 +45,9 @@ var (
 	resyncPeriod = flag.Duration("resync-period", time.Minute*10, "Resync period for cache")
 	workers      = flag.Int("workers", 10, "Concurrency to process multiple resize requests")
 
-	csiAddress  = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
-	csiTimeout  = flag.Duration("csiTimeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
+	csiAddress = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	timeout    = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+
 	showVersion = flag.Bool("version", false, "Show version")
 
 	retryIntervalStart = flag.Duration("retry-interval-start", time.Second, "Initial retry interval of failed volume resize. It exponentially increases with each failure, up to retry-interval-max.")
@@ -99,7 +101,7 @@ func main() {
 
 	csiResizer, err := resizer.NewResizer(
 		*csiAddress,
-		*csiTimeout,
+		*timeout,
 		kubeClient,
 		informerFactory,
 		*metricsAddress,
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Note that the external-resizer does not scale with more replicas. Only one exter
 
 * `--leader-election-namespace`: Namespace where the leader election resource lives. Defaults to the pod namespace if not set.
 
-* `--csiTimeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerExpandVolume` calls. 15 seconds is used by default.
+* `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerExpandVolume` calls. 10 seconds is used by default.
 
 * `--retry-interval-start`: The starting value of the exponential backoff for failures. 1 second is used by default.
 


Madhu-1:
Fixes https://github.com/kubernetes-csi/external-resizer/issues/72

cmd/csi-resizer/main.go
```diff
@@ -41,7 +40,7 @@ var (
 	workers      = flag.Int("workers", 10, "Concurrency to process multiple resize requests")
 
 	csiAddress  = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
-	csiTimeout  = flag.Duration("csiTimeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
+	timeout     = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
```
Madhu-1, line 43: IMO we should not remove the parameter `csiTimeout`, we should deprecate old parameter and add a new one first

humblec: Forgot that we havent called out for deprecation yet, will change accordingly.

gnufied:
@humblec can you update this PR and rebase it?

humblec:
> @humblec can you update this PR and rebase it?

Sure. rebased @gnufied . PTAL.

cmd/csi-resizer/main.go
```diff
@@ -99,7 +102,7 @@ func main() {
 
 	csiResizer, err := resizer.NewResizer(
 		*csiAddress,
-		*csiTimeout,
+		*timeout,
```
gnufied, line 105: this means that if someone was using deprecated parameter - `csiTimeout`, it will simply be ignored. Can you fix it?

gnufied: Since in this release we are only deprecating the older `csiTimeout` parameter (as evident from CLI help flags), but if we remove the flag in the same release we will have problems.

gnufied: alternately - if we are not going to support `csiTimeout` flag anymore, we should remove it alogether. But I am in favour of deprecating the flag for now and removing it later. cc @msau42

msau42: If someone sets csiTimeout, can we have it set timeout?

I'm also ok with just removing the flag because we're doing a major release anyway.

humblec: Ok.. let me try to explain what I was thinking. Being this is a major release, we could definitely mark `csitimeout` param as a "deprecated" one , also we could mention in the release note that if someone set this value , it will be NO-OP and the default timeout value in the external resizer will be used. The advantage with this approach is that, the older templates ( which set `csiTimeout` ) in some customer/user setup could also work , but its a NO-OP . If they set `timeout` while using new templates in parity with this release, it will be effective as well. Isnt it a good idea to address this way? @gnufied @msau42 wdyt?

gnufied: that would be problematic. It would seem like  we still support `csiTimeout` option but in truth we don't. If we are going to not support `csiTimeout` option then we should just remove it.

humblec: @gnufied @msau42 ok.. are we in agreement to completely remove `csiTimeout` ?

gnufied: yep.

humblec: @gnufied ptal on current version.. completely removed `csiTimeout`.

gnufied:
@humblec can you add a release note plz?

humblec:
@gnufied does this looks good as release note for this?

```

The `--csiTimeout` flag of external-resizer is obsoleted with this release. Instead `--timeout` flag has been introduced for the same purpose. The   timeout value  accommodates majority of `ControllerExpandVolume` calls. `timeout` value defaults to 15 seconds. From now on, the deployment  templates or  artifacts has to make use of `--timeout` flag instead of obsoleted `--csiTimeout`  flag.
```

cmd/csi-resizer/main.go
```diff
@@ -44,8 +45,9 @@ var (
 	resyncPeriod = flag.Duration("resync-period", time.Minute*10, "Resync period for cache")
 	workers      = flag.Int("workers", 10, "Concurrency to process multiple resize requests")
 
-	csiAddress  = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
-	csiTimeout  = flag.Duration("csiTimeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
+	csiAddress = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	timeout    = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
```
msau42, line 49: How come 15? All the other sidecars are 10 seconds. Actually, do we want to default a timeout value? I find 10 seconds is quite short for real drivers. xref: https://github.com/kubernetes-csi/external-provisioner/issues/451#issuecomment-662604169

humblec: @msau42 sure we can keep in parity between sidecars ie 10 seconds.  I kept it 15 seconds as obsoleted  `csitimeout ` default was "15s".

humblec: ok.. I changed to 10s default @msau42 @gnufied

cmd/csi-resizer/main.go
gnufied: hmm, I think what @msau42 meant was 15s itself is too low of a default (and hence 10s will be even lower) and hence is it worth forcing COs/cluster-admins to specify a value rather than choosing a non-ideal default?

I see @msau42 point but I kinda like the idea of defaults, but I admit 15s is not good default for volume expansion. I wonder if we can increase the default to something like 80s which would work in most cases and if certain drivers have problems they can tune it.  

Also this PR will require fixing all our e2e jobs btw.

cmd/csi-resizer/main.go
gnufied: Also - I kind of dislike, renaming the parameter and making it required in same release. It causes problems for downstream users. So, if we choose to make this parameter mandatory, we should do this in 2.0 IMO. 

@msau42 does 80s as a default sound good to you? The reason I picked this is because, from my experience this is "good enough" default for most in-tree plugins for which there is a equivalent CSI driver. If we all agree @humblec can you update the PR to use `80s` as default timeout?

humblec: sure @gnufied , so once @msau42 ack on `80s` , I will change the default value to the same :+1:

humblec: Also, I do second to `2.0.0` with this new parameter instead of changing it in the minor release.

gnufied: we spoke in CSI implementation call today. It was decided that, 10s is a good default because it exposes driver idempotency issues and specific drivers can tune this value.

humblec: Thanks for the update @gnufied ! :+1:

cmd/csi-resizer/main.go
```diff
@@ -44,8 +45,9 @@ var (
 	resyncPeriod = flag.Duration("resync-period", time.Minute*10, "Resync period for cache")
 	workers      = flag.Int("workers", 10, "Concurrency to process multiple resize requests")
 
-	csiAddress  = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
-	csiTimeout  = flag.Duration("csiTimeout", 15*time.Second, "Timeout for waiting for CSI driver socket.")
+	csiAddress = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
```
msau42, line 48: Since we're going to ignore this csiAddress flag, should we use the [deprecatedflag library](https://github.com/kubernetes-csi/csi-lib-utils/tree/master/deprecatedflags)?

gnufied: wait, why is `csiAddress` being ignored?

gnufied: sorry this got merged before this has chance to get addressed.  but we can do a follow up.

Issue #81, opened 2020-05-14
Respect volume-in-use error when calling ControllerExpand volume


gnufied:
if a CSI driver throws volume-in-use error when calling ControllerExpandVolume, the external-resizer should not retry expansion until it can verify that volume is not in use. This will be a best effort check and only be performed after a plugin has thrown `Volume in use` error. This is different from enforcing online and offline plugin capabilities.

cc https://github.com/kubernetes-csi/external-resizer/issues/62

gnufied:
cc @xing-yang @msau42 

/assign

PR #86, opened 2020-06-03
Prevent calling controller-expand if volume in-use

gnufied:
If volume can not be expanded when in-use we must not call ControllerExpandVolume until we know volume is not in-use.

Fixes https://github.com/kubernetes-csi/external-resizer/issues/81

```release-note
Do not call controller-expand if volume is in-use and can not be expanded while in-use
```


pkg/controller/controller.go
```diff
@@ -98,17 +112,38 @@ func NewResizeController(
...
+	objKey, err := getObjectKey(obj)
 	if err != nil {
 		return
 	}
 	ctrl.claimQueue.Add(objKey)
 }
 
+func (ctrl *resizeController) addPod(obj interface{}) {
+	pod := parsePod(obj)
+	if pod != nil {
```
xing-yang, line 133: Should this be "if pod == nil"?  Otherwise pod is nil on line 136.

pkg/controller/controller.go
```diff
@@ -98,17 +112,38 @@ func NewResizeController(
...
+		return
+	}
+	ctrl.usedPVCs.addPod(pod)
+}
+
+func (ctrl *resizeController) deletePod(obj interface{}) {
+	pod := parsePod(obj)
+	if pod != nil {
```
xing-yang, line 141: Same comment here.

pkg/controller/controller.go
```diff
@@ -98,17 +112,38 @@ func NewResizeController(
 		DeleteFunc: ctrl.deletePVC,
 	}, resyncPeriod)
 
+	podInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
```
xing-yang, line 115: I see that the in-tree PVC protection controller also checks the pod.  What is the advantage of checking thru the pod vs checking the VolumeAttachment object?

pkg/controller/controller.go
```diff
@@ -322,6 +357,15 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 		pvc = updatedPVC
 	}
 
+	// if pvc previously failed to expand because it can't be expanded when in-use
+	// we must not try expansion here
+	if ctrl.usedPVCs.hasInUseErrors(pvc) && ctrl.usedPVCs.checkForUse(pvc) {
+		// Record an event to indicate that resizer is not expanding the pvc
+		ctrl.eventRecorder.Event(pvc, v1.EventTypeWarning, util.VolumeResizeFailed,
+			fmt.Sprintf("CSI resizer is not expanding %s because it is in-use", pv.Name))
```
xing-yang, line 365: Can you add a message to indicate that the CSI driver only supports offline volume expansion.

pkg/controller/controller.go
```diff
@@ -422,3 +478,37 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
 	return nil
 }
+
+func parsePod(obj interface{}) *v1.Pod {
+	if obj == nil {
+		return nil
+	}
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
```
xing-yang, line 488: Can you pick a different name than "tombstone"?

pkg/controller/controller.go
```diff
@@ -422,3 +478,37 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
+}
+
+func inUseError(err error) bool {
+	st, ok := status.FromError(err)
+	if !ok {
+		// not a grpc error
+		return false
+	}
+	// if this is a failed precondition error then that means driver does not support expansion
+	// of in-use volumes
```
xing-yang, line 509: Can you add a comment linking to the CSI spec description about this?

pkg/controller/controller.go
gnufied: good catch. thank you.

pkg/controller/controller.go
gnufied: Because not all volume types support attach/detach and hence there may not be an VolumeAttachment object.

pkg/controller/controller.go
gnufied: done

pkg/controller/controller.go
gnufied: this is fixed. Also added tests that cover this behaviour.

pkg/controller/controller.go
gnufied: done.

pkg/controller/controller.go
gnufied: renamed to `staleObj`

pkg/controller/controller.go
```diff
@@ -322,6 +367,15 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 		pvc = updatedPVC
 	}
 
+	// if pvc previously failed to expand because it can't be expanded when in-use
+	// we must not try expansion here
+	if ctrl.usedPVCs.hasInUseErrors(pvc) && ctrl.usedPVCs.checkForUse(pvc) {
+		// Record an event to indicate that resizer is not expanding the pvc
+		msg := fmt.Sprintf("Unable to expand %s because CSI driver %s only supports offline expansion and volume is currently in-use", pv.Name, ctrl.resizer.Name())
```
xing-yang, line 374: I think print out namespace/name of the PVC is better than PV name in the event.

pkg/controller/controller.go
```diff
@@ -422,3 +488,38 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
+
+func parsePod(obj interface{}) *v1.Pod {
+	if obj == nil {
+		return nil
+	}
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		staleObj, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("couldn't get object from tombstone %#v", obj))
```
xing-yang, line 500: s/tombstone/staleObj

pkg/controller/controller.go
```diff
@@ -422,3 +488,38 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
+			return nil
+		}
+		pod, ok = staleObj.Obj.(*v1.Pod)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a Pod %#v", obj))
```
xing-yang, line 505: s/tombstone/staleObj

PR #96, opened 2020-08-10
We need to ensure that we are not reading stale PV objects

gnufied:
So as we don't end up resizing same object more than once

Fixes flakes associated with resizing.

/kind bug

```release-note
Ensure that we do not resize recently resized volumes
```


gnufied:
This fixes flake coming from - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-resizer/82/pull-kubernetes-csi-external-resizer-1-16-on-kubernetes-1-16/1292898043053477890

pkg/controller/controller.go
```diff
@@ -341,6 +341,15 @@ func (ctrl *resizeController) pvNeedResize(pvc *v1.PersistentVolumeClaim, pv *v1
 		return false
 	}
 
+	// lets make sure that we are reading latest PV object so as any PV updates that might have happened
```
msau42, line 344: What happens if you call resize twice on the same PV?

gnufied: in general once kubelet finishes resizing then all resizing related conditions are removed from PVC. But what is happening in e2e is - we take down the pod and then we expand the PVC and we wait for "FileSystemResizePending" condition on PVC, but if resize is called again in resize controller then that condition might get replaced by "Resizing" condition and hence e2e may flake.

msau42: Does that mean we should keep an internal cache and update it after we make an API update, similar to what we do in pv controller? That way, we don't need to do a GET request on every loop, which will limit our scalibility.

gnufied: Also the reason I think this will eliminate this kind of flakes is because - the previous operation on PVC can't finish without writing the object to etcd. The only reason we would return true from here is if informers haven't been updated yet, but if we did read from apiserver directly, this problem should go away.

Alternatively we will have to maintain local cache like provisioner/snapshotter etc.

gnufied: I was thinking about local caching but I am not sure if it is worth doing in this case. The PV size comparison is always done *after* PVC size comparsion, so in this case PVC does have a volume expansion pending (like we know `pvc.Status.Capacity < pvc.Spec.Capacity`). So this API call should only be made twice per PVC size change.

msau42: cc @RaunakShah. Can you run your benchmark tests against this PR so we can see the impact of adding this GET call?

RaunakShah: Sure!

RaunakShah: Spawning 10 threads to resize PVCs in a loop for 120s. Each thread moves onto the next resize operation once the PVC enters `FileSystemResizePending` condition. Dummy vSphere CSI driver that returns success for `ControllerExpandVolume` calls. 
- Without this change, using image `quay.io/k8scsi/csi-resizer:v1.0.0-rc2`:
Ran 134 operations with throughput of 1.127 ops/sec
- With this change:
Ran 88 operations with a throughput of 0.79 ops/sec

gnufied: So if I understand correctly, this PR overall reduced number of API requests from external-resizer?

RaunakShah: Quite the opposite unfortunately :) With the rc image, i can get 134 calls to `ControllerExpandVolume` but that drops to 88 with this PR, the reason being that each `ControllerExpandVolume` request takes longer to process.

gnufied: @jsafrane pointed out that we could do same thing @RaunakShah  did in provisioner - https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/83 .  lets see if this helps.

gnufied:
@msau42 can you PTAL. I have moved the PR to use `cache.Store` object and hence should not incur API cost.

pkg/controller/controller.go
```diff
@@ -282,22 +285,30 @@ func (ctrl *resizeController) syncPVC(key string) error {
...
 	}
 
+	if !exists {
+		klog.V(3).Infof("PVC %s/%s is deleted or does not exist", namespace, name)
+		return nil
+	}
+
+	pvc, ok := pvcObject.(*v1.PersistentVolumeClaim)
+	if !ok {
+		klog.Errorf("expected PVC got: %v", pvcObject)
```
msau42, line 302: Can we let the caller log the error? Ditto throughout

pkg/controller/controller.go
```diff
@@ -407,7 +434,7 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 		ctrl.eventRecorder.Eventf(pvc, v1.EventTypeWarning, util.VolumeResizeFailed, err.Error())
 	}
 
-	return err
+	return ctrl.claims.Update(updatedPVC)
```
msau42, line 437: We already update it in `resizeVolume`. Do we need to update it here too? Would it be simpler if we just always update it after immediately doing a successful api update?

gnufied: We did not update PVC in `resizeVolume` function. We updated PV in there.

pkg/controller/controller.go
gnufied: fixed.

pkg/controller/controller.go
```diff
@@ -307,6 +318,16 @@ func (ctrl *resizeController) syncPVC(key string) error {
 		return err
 	}
 
+	if !exists {
```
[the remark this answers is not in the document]
gnufied: fixed.

pkg/controller/controller.go
msau42: Same comment applies. Can we update the cache immediately after the update/patch operation, instead of the callers needing to realize that the function updated the api?

pkg/controller/controller.go
```diff
@@ -375,6 +387,12 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 		pvc = updatedPVC
 	}
 
+	err := ctrl.claims.Update(pvc)
+	if err != nil {
+		klog.Errorf("error updating PVC %s in cache: %v", util.PVCKey(pvc), err)
```
msau42, line 392: Can we return this error and let the caller log?

pkg/controller/controller.go
```diff
@@ -435,11 +453,17 @@ func (ctrl *resizeController) resizeVolume(
...
+	updatedPV, err := util.UpdatePVCapacity(pv, newSize, ctrl.kubeClient)
 
-	if err := util.UpdatePVCapacity(pv, newSize, ctrl.kubeClient); err != nil {
+	if err != nil {
 		klog.Errorf("Update capacity of PV %q to %s failed: %v", pv.Name, newSize.String(), err)
 		return newSize, fsResizeRequired, err
 	}
+	err = ctrl.volumes.Update(updatedPV)
+	if err != nil {
+		klog.Errorf("error updating PV %s: %v", updatedPV.Name, err)
```
msau42, line 464: Can we return the error and let the caller log?

pkg/controller/controller.go
```diff
@@ -460,22 +484,23 @@ func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeCl
...
 	pvc *v1.PersistentVolumeClaim,
-	newSize resource.Quantity) error {
+	newSize resource.Quantity) (*v1.PersistentVolumeClaim, error) {
 	newPVC := pvc.DeepCopy()
 	newPVC.Status.Capacity[v1.ResourceStorage] = newSize
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(pvc.Status.Conditions, []v1.PersistentVolumeClaimCondition{})
-	if _, err := util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient); err != nil {
+	updatedPVC, err := util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient)
+	if err != nil {
 		klog.Errorf("Mark PVC %q as resize finished failed: %v", util.PVCKey(pvc), err)
```
msau42, line 493: return error

pkg/controller/controller.go
```diff
@@ -486,15 +511,18 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions,
 		[]v1.PersistentVolumeClaimCondition{pvcCondition})
 
-	if _, err := util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient); err != nil {
+	updatedPVC, err := util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient)
+
+	if err != nil {
 		klog.Errorf("Mark PVC %q as file system resize required failed: %v", util.PVCKey(pvc), err)
```
msau42, line 517: return error

pkg/controller/controller.go
gnufied: fixed.

pkg/controller/controller.go
gnufied: fixed

pkg/controller/controller.go
gnufied: I have moved code that updates the cache to be in same place where PV or PVC object is updated.

pkg/controller/controller.go
gnufied: done.

pkg/controller/controller.go
```diff
@@ -406,8 +417,7 @@ func (ctrl *resizeController) resizePVC(pvc *v1.PersistentVolumeClaim, pv *v1.Pe
 		// Record an event to indicate that resize operation is failed.
 		ctrl.eventRecorder.Eventf(pvc, v1.EventTypeWarning, util.VolumeResizeFailed, err.Error())
 	}
-
-	return err
+	return nil
```
msau42, line 420: Why is this changed to nil?

pkg/controller/controller.go
```diff
@@ -426,19 +436,22 @@ func (ctrl *resizeController) resizeVolume(
...
 	klog.V(4).Infof("Resize volume succeeded for volume %q, start to update PV's capacity", pv.Name)
+	updatedPV, err := util.UpdatePVCapacity(pv, newSize, ctrl.kubeClient)
 
-	if err := util.UpdatePVCapacity(pv, newSize, ctrl.kubeClient); err != nil {
-		klog.Errorf("Update capacity of PV %q to %s failed: %v", pv.Name, newSize.String(), err)
-		return newSize, fsResizeRequired, err
+	if err != nil {
+		return newSize, fsResizeRequired, fmt.Errorf("Update capacity of PV %q to %s failed: %v", pv.Name, newSize.String(), err)
+	}
+	err = ctrl.volumes.Update(updatedPV)
```
msau42, line 452: I was thinking to do the cache update in the util method itself, but I guess we don't want to pass `ctrl` in. Can we create another helper function here?

pkg/controller/controller.go
```diff
@@ -426,19 +436,22 @@ func (ctrl *resizeController) resizeVolume(
...
 
 	if err != nil {
-		klog.Errorf("Resize volume %q by resizer %q failed: %v", pv.Name, ctrl.name, err)
 		// if this error was a in-use error then it must be tracked so as we don't retry without
 		// first verifying if volume is in-use
 		if inUseError(err) {
 			ctrl.usedPVCs.addPVCWithInUseError(pvc)
 		}
-		return newSize, fsResizeRequired, fmt.Errorf("resize volume %s failed: %v", pv.Name, err)
+		return newSize, fsResizeRequired, fmt.Errorf("Resize volume %q by resizer %q failed: %v", pv.Name, ctrl.name, err)
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/controller/controller.go
gnufied: not sure how this happeend. Added a unit test to ensure this does not happen. good catch.

pkg/controller/controller.go
gnufied: made a new helper function that updates the PV with api-server and updates local cache.

pkg/controller/controller.go
```diff
@@ -455,7 +464,15 @@ func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeCl
...
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions,
 		[]v1.PersistentVolumeClaimCondition{progressCondition})
-	return util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient)
+	updatedPVC, err := util.PatchPVCStatus(pvc, newPVC, ctrl.kubeClient)
+	if err != nil {
+		return nil, err
+	}
+	err = ctrl.claims.Update(updatedPVC)
+	if err != nil {
+		return nil, fmt.Errorf("error updating PVC %s in local cache: %v", util.PVCKey(newPVC), err)
```
RaunakShah, line 473: Do we want to fail the entire operation if the local cache fails to update?

gnufied: In the worst case when updating local cache failed and the informer cache did not update itself (from PVC events) while operation was retried - the test will flake.

So, it should be fine I think.

pkg/controller/controller.go
gnufied: We do have now helpers functions on controller type which do the same. They patch the object and then they update the cache. Functions - `updatePVCapacity`, `markPVCAsFSResizeRequired` and `markPVCResizeFinished` both update the object with apiserver and then update the object in the cache.

Did you wanted both patching and updating in cache to be done within util functions which aren't part of controller type? In that case - we will have to pass controller instance to those functions and in which case these functions may as well be on controller type (which is what I have done).

gnufied: I was thinking about doing that but that means `PatchPVCStatus` function can't really belong to util package because it has to also update the controller's cache. I hear what you are saying but I am not sure if it is worth doing given the size of resize controller which has pretty small surface area of PV and PVC updates.

PR #129, opened 2020-12-05
Bumping csi-lib-utils to 0.9.0

verult:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

```release-note
* Added leader election health check at the metrics port + path "/healthz/leader-election".
* klog/v2 is used for logging
* process_start_time metric is now reported
* `metrics-address` flag is deprecated and replaced by `http-endpoint`, which enables handlers from both metrics manager and leader election health check.
```

/assign @msau42


cmd/csi-resizer/main.go
```diff
@@ -99,17 +110,29 @@ func main() {
...
+		addr,
 		*metricsPath)
 	if err != nil {
 		klog.Fatal(err.Error())
 	}
 
+	// Start HTTP server for metrics + leader election healthz
+	go func() {
+		klog.Infof("ServeMux listening at %q", addr)
+		err := http.ListenAndServe(addr, mux)
```
msau42, line 130: do we need to check addr empty?

pkg/resizer/csi_resizer.go
```diff
@@ -68,15 +70,18 @@ func NewResizerFromClient(
...
 	metricsAddress, metricsPath string) (Resizer, error) {
 	driverName, err := getDriverName(csiClient, timeout)
 	if err != nil {
 		return nil, fmt.Errorf("get driver name failed: %v", err)
 	}
 
 	klog.V(2).Infof("CSI driver name: %q", driverName)
-	metricsManager.SetDriverName(driverName)
-	metricsManager.StartMetricsEndpoint(metricsAddress, metricsPath)
+	if metricsAddress != "" {
```
msau42, line 81: is it possible to move this outside so that it's similar to how csi-attacher does it?

cmd/csi-resizer/main.go
verult: thanks for the catch!

pkg/resizer/csi_resizer.go
verult: I added the refactor as a separate commit. WDYT?

PR #140, opened 2021-02-23
fix: use pv annotation to trigger filesystem resize when necessary

sunpa93:
**What type of PR is this?**
 /kind bug

fix: add annotation to pv after resize if filesystem resize is necessary. delete when pvc.status.capacity >= pv.spec.capacity

Fixes https://github.com/kubernetes/kubernetes/issues/88683

This PR is a sister PR of https://github.com/kubernetes/kubernetes/pull/99326

```release-note
NONE
```


pkg/controller/controller.go
```diff
@@ -533,11 +551,34 @@ func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClai
...
+}
+
+func (ctrl *resizeController) updatePVCapacity(pv *v1.PersistentVolume, oldCapacity, newCapacity resource.Quantity, fsResizeRequired bool) error {
 	klog.V(4).Infof("Resize volume succeeded for volume %q, start to update PV's capacity", pv.Name)
 	newPV := pv.DeepCopy()
 	newPV.Spec.Capacity[v1.ResourceStorage] = newCapacity
 
+	if fsResizeRequired {
+		// if the pv already has a resize annotation skip
+		if metav1.HasAnnotation(pv.ObjectMeta, util.AnnPreResizeCapacity) {
```
gnufied, line 573: I think this is problematic - if I expand a PV from 10 to 20GB and while expansion is pending on the node, I expand it again to 40GB then second expansion won't be recorded in the pv spec?

pkg/controller/controller.go
```diff
@@ -533,11 +551,34 @@ func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClai
...
+			return nil
+		}
+		if newPV.ObjectMeta.Annotations == nil {
+			newPV.ObjectMeta.Annotations = make(map[string]string)
+		}
+		newPV.ObjectMeta.Annotations[util.AnnPreResizeCapacity] = (&oldCapacity).String()
```
gnufied, line 579: Why are we doing - `(&oldCapacity).String()` is that different from `oldCapacity.String()` ?

gnufied (commented):
Looks good mostly. But I think we need unit tests for some these new behaviours.

sunpa93:
> Looks good mostly. But I think we need unit tests for some these new behaviours.

I have modified the tests so that they check if annotations get properly populated and deleted

pkg/controller/controller.go
```diff
@@ -320,6 +321,16 @@ func (ctrl *resizeController) syncPVC(key string) error {
 		return fmt.Errorf("expected volume but got %+v", volumeObj)
 	}
 
+	if ctrl.isFsResizeComplete(pvc, pv) && metav1.HasAnnotation(pv.ObjectMeta, util.AnnPreResizeCapacity) {
```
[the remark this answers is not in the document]
sunpa93: I have renamed the function. I noticed that there are several places in the code that uses fsResize word. Will there be future issues to address this refactoring?

pkg/controller/controller.go
```diff
@@ -320,6 +321,16 @@ func (ctrl *resizeController) syncPVC(key string) error {
...
+		if err := ctrl.deletePreResizeCapAnnotation(pv); err != nil {
+			return fmt.Errorf("Failed removing annotation %s from pv %q: %v", util.AnnPreResizeCapacity, pv.Name, err)
```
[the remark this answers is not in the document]
sunpa93: Done :)

pkg/controller/controller.go
sunpa93: You are definitely right on this. The function should not return before updating `pv.Spec.Capacity`

pkg/controller/controller.go
sunpa93: Done :)

pkg/controller/controller.go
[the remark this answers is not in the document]
gnufied: There is still conditions on the PVC which use fsResize wording, we can't rename that because that has become part of API now. But other than that we could and should rename everything else.

pkg/controller/controller.go
```diff
@@ -320,6 +321,16 @@ func (ctrl *resizeController) syncPVC(key string) error {
 		return fmt.Errorf("expected volume but got %+v", volumeObj)
 	}
 
+	if ctrl.isNodeExpandComplete(pvc, pv) && metav1.HasAnnotation(pv.ObjectMeta, util.AnnPreResizeCapacity) {
+		if err := ctrl.deletePreResizeCapAnnotation(pv); err != nil {
```
msau42, line 325: how come resize controller is deleting the annotation? Shouldn't kubelet do it after nodeexpand is complete?

sunpa93: This is because kubelet does not have the permission to modify PV objects.

cmd/csi-resizer/main.go
```diff
@@ -68,6 +68,8 @@ var (
 
 	handleVolumeInUseError = flag.Bool("handle-volume-inuse-error", true, "Flag to turn on/off capability to handle volume in use error in resizer controller. Defaults to true if not set.")
 
+	enableFSResizeAnnotation = flag.Bool("enable-fs-resize-annotation", true, "Flag to turn on/off capability to add an annotation to PV to denote its size prior to resize and to remove it upon completion of filesystem resize.")
+
```
gnufied, line 72: I think we wanted this as a feature gate and not a flag.  The k8s feature gates look like - `--feature-gate=RestoreDeletedVolumeSize`

msau42: You can look at external-provisioner feature gate handling.

pkg/controller/controller.go
msau42: @gnufied is this an issue? Does this mean plugins that don't support controller resize and don't run the resizer will get this fix?

Jiawei0227:
Any updates on this? We are pending on csi-resizer release

cmd/csi-resizer/main.go
sunpa93: Done :)

pkg/controller/controller.go
```diff
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	features "sigs.k8s.io/external-resizer/pkg/features" /* TO BE REPLACED WITH features "github.com/external-resizer/pkg/features once PR merged*/
```
gnufied, line 24: Why are we importing via `sigs.k8s.io`? I do not understand reason for doing this. you should be able to import it normally via `github.com/kubernetes-csi/external-resizer`.

pkg/controller/controller.go
gnufied: Currently I think external-resizer is required even for plugins that don't support controller expansion, because in that case external-resizer calls a NO-OP control-plane expansion and successfully expands the PV. There is no way to update PV spec from kubelet otherwise. So every plugin that supports expansion needs to run resizer.

pkg/controller/controller.go
sunpa93: Done 👍

gnufied:
Can you squash 4 last commits in one?

sunpa93:
> Can you squash 4 last commits in one?

Done :)

PR #155, opened 2021-06-02
Reword file system expansion wording to node expansion

gnufied:
This PR does not changes event keys and neither we are changing
condition reported in the PVC.

cc @msau42 

```release-note
Rename file system resizing events to node expansion required
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -481,7 +481,7 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
 		Type:               v1.PersistentVolumeClaimFileSystemResizePending,
 		Status:             v1.ConditionTrue,
 		LastTransitionTime: metav1.Now(),
-		Message:            "Waiting for user to (re-)start a pod to finish file system resize of volume on node.",
+		Message:            "Waiting for user to (re-)start a pod to finish expansion of volume on the node.",
 	}
 	newPVC := pvc.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -490,12 +490,12 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
...
 	if err != nil {
-		return fmt.Errorf("Mark PVC %q as file system resize required failed: %v", util.PVCKey(pvc), err)
+		return fmt.Errorf("Failed to mark PVC %q for node expansion: %v", util.PVCKey(pvc), err)
 	}
 
-	klog.V(4).Infof("Mark PVC %q as file system resize required", util.PVCKey(pvc))
+	klog.V(4).Infof("Mark PVC %q for required node expansion", util.PVCKey(pvc))
 	ctrl.eventRecorder.Eventf(pvc, v1.EventTypeNormal,
-		util.FileSystemResizeRequired, "Require file system resize of volume on node")
+		util.FileSystemResizeRequired, "Require expansion of volume on node")
```
gnufied, line 498: I wonder if we can change the event keys though. The underlying motivation for this PR is - for raw block devices etc, they may not even need a file system expansion (but may still need rescannig the LUN on the node). So it is bit broad to call node expansion as a file system expansion.

gnufied: I renamed these keys. I think nobody should generally depend on the event keys.

gnufied:
The failing test looks unrelated.

PR #178, opened 2021-10-19
Fix: PVC used by a job doesn't get resize after the pod of the job completed

### LINKED ISSUE #175: PVC used by a job doesn't get resize after the pod of the job completed
**Summary:**
We have a setup in which the `external-resizer` is used with the storage provider that only supports offline expansion (e.g., only supports `PluginCapability_VolumeExpansion_OFFLINE`).  We deployed a job that uses a PVC provisioned by the storage provider. While the job pod is running, we resize the PVC by modifying `spec.resources.requests.storage`. The PVC cannot be resized while the pod is running as expected. However, after the job pod is completed, the PVC still doesn't get resized. `external-resizer`doesn't send resizing gRPC call to the storage provider. The PVC is stuck in this state forever until we manually delete the job pod.

**Reproduce steps:**
1. Deploy `external-resizer` together with a storage provider (we use [Longhorn](https://github.com/longhorn/longhorn))
2. Don't set the `--handle-volume-inuse-error` flag for the `external-resizer` . It means that by default, `external-resizer`  will handle handle volume in use error in resizer controller, [link](https://github.com/kubernetes-csi/external-resizer/blob/45b15b7a829990e37fb4d5760b96de2772dc74ee/cmd/csi-resizer/main.go#L73)
3. Deploy a job that uses a PVC as below. The job creates a pod that will sleep for 2 minutes and complete.
    <details>
    <summary>Click to open</summary>
    
    ```
    apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      name: test-job-pvc
      namespace: default
    spec:
      accessModes:
        - ReadWriteOnce
      storageClassName: longhorn
      resources:
        requests:
          storage: 1Gi
    ---
    apiVersion: batch/v1
    kind: Job
    metadata:
      name: test-job
      namespace: default
    spec:
      backoffLimit: 1
      template:
        metadata:
          name: test-job
        spec:
          containers:
            - name: test-job
              image: ubuntu:latest
              imagePullPolicy: IfNotPresent
              securityContext:
                privileged: true
              command: ["/bin/sh"]
              args: ["-c", "echo 'sleep for 120s then exit'; sleep 120"]
              volumeMounts:
                - mountPath: /data
                  name: vol
          restartPolicy: OnFailure
          volumes:
            - name: vol
              persistentVolumeClaim:
                claimName: test-job-pvc
    ```
    </details>
4. While the job pod become running, try to expand the PVC by editing the `spec.resources.requests.storage`
5. Observe that the resizing fail
6. Wait for the job pod to become completed.
7. Observer that that PVC stuck in the current state forever. It doesn't get resized because `external-resizer` doesn't attempt to make gRPC expanding call to the storage provider.
 
**Expected Behavior:**

Once the job pod is completed, the PVC is no longer consider to be `in-used`. Therefore  `external-resizer` should attempt to make gRPC expanding call to the storage provider.

**Propose:**
We dig into the source code see that:
1. This [checker](https://github.com/kubernetes-csi/external-resizer/blob/45b15b7a829990e37fb4d5760b96de2772dc74ee/pkg/controller/controller.go#L413) prevent the  `external-resizer` from retrying if the PVC has InUseErrors before AND it is in the `ctrl.usedPVCs` map
2. The problem is that the PVC is never removed from the `ctrl.usedPVCs` map when a pod move to `completed` phase. PVC is only removed when the pod is deleted, [link](https://github.com/kubernetes-csi/external-resizer/blob/45b15b7a829990e37fb4d5760b96de2772dc74ee/pkg/controller/controller.go#L150)
3. We think that the logic [over here](https://github.com/kubernetes-csi/external-resizer/blob/45b15b7a829990e37fb4d5760b96de2772dc74ee/pkg/controller/controller.go#L158) should be changed to handle the case when the pod become `completed`. I.e.,:
    ```
    func (ctrl *resizeController) updatePod(oldObj, newObj interface{}) {
	    pod := parsePod(newObj)
	    if pod == nil {
		    return
	    }
	    
	    if isPodTerminated(pod) {
		    ctrl.usedPVCs.removePod(pod)
	    } else {
		    ctrl.usedPVCs.addPod(pod)
	    }
    }
    ```
 
**Evn:**
* `external-resizer` v1.2.0
* [Longhorn](https://github.com/longhorn/longhorn) v1.2.2

PhanLe1010:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #175

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -161,7 +161,11 @@ func (ctrl *resizeController) updatePod(oldObj, newObj interface{}) {
 		return
 	}
 
-	ctrl.usedPVCs.addPod(pod)
+	if isPodTerminated(pod) {
+		ctrl.usedPVCs.removePod(pod)
+	} else {
+		ctrl.usedPVCs.addPod(pod)
+	}
 }
 
 func (ctrl *resizeController) updatePVC(oldObj, newObj interface{}) {


PhanLe1010:
Test passed. ping @gnufied

PR #184, opened 2021-12-13
update kube and other dependencies to v1.23

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

```release-note
Updates Kubernetes dependencies to v1.23.0
```


gnufied:
@humblec actually can you update the release-note to say client-go etc were updated? We seem to be capturing that information in changelogs traditionally.

PR #187, opened 2022-01-13
Handle case of recovery from resize failures

gnufied:
Handle the case of recovery from resize failure.

xref https://github.com/kubernetes/enhancements/issues/1790

This is mostly same implementation from intree controller code - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/operationexecutor/operation_generator.go#L1782 with minor changes because of node-only expansion handling as described in KEP.

Manually tested following scenarios (with hostpath driver):
- [x] If controller-expansion failed, it allows recovery to a size > pvc.Status.Cap
- [x] if Node expansion is pending, it does not try expansion to new size until previous expansion is finished
- [x] Tested with consecutive changes which can cause resize success and failure
- [x] Tested with node expansion failing (control-plane succeeded) and users reducing size (no recovery is possible)
- [x] Tested with node expansion failing but no controller-expansion is available (recovery is possible)

```release-note
Add support to allow users to recover from expansion failures
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -343,12 +343,18 @@ func (ctrl *resizeController) syncPVC(key string) error {
 		klog.V(4).Infof("No need to resize PVC %q", util.PVCKey(pvc))
 		return nil
 	}
-	if !ctrl.pvNeedResize(pvc, pv) {
-		klog.V(4).Infof("No need to resize PV %q", pv.Name)
-		return nil
-	}
 
-	return ctrl.resizePVC(pvc, pv)
+	if utilfeature.DefaultFeatureGate.Enabled(features.RecoverVolumeExpansionFailure) {
+		_, _, err, _ := ctrl.expandAndRecover(pvc, pv)
+		return err
+	} else {
+		if !ctrl.pvNeedResize(pvc, pv) {
+			klog.V(4).Infof("No need to resize PV %q", pv.Name)
+			return nil
+		}
+
+		return ctrl.resizePVC(pvc, pv)
+	}
 }
 
 // pvcNeedResize returns true is a pvc requests a resize operation.
@@ -476,7 +482,7 @@ func (ctrl *resizeController) resizeVolume(
 	}
 	klog.V(4).Infof("Resize volume succeeded for volume %q, start to update PV's capacity", pv.Name)
 
-	err = ctrl.updatePVCapacity(pv, pvc.Status.Capacity[v1.ResourceStorage], newSize, fsResizeRequired)
+	_, err = ctrl.updatePVCapacity(pv, pvc.Status.Capacity[v1.ResourceStorage], newSize, fsResizeRequired)
 	if err != nil {
 		return newSize, fsResizeRequired, err
 	}
@@ -496,7 +502,7 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions,
 		[]v1.PersistentVolumeClaimCondition{pvcCondition})
 
-	_, err := ctrl.patchClaim(pvc, newPVC)
+	_, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
 
 	if err != nil {
 		return fmt.Errorf("Mark PVC %q as file system resize required failed: %v", util.PVCKey(pvc), err)
@@ -509,6 +515,7 @@ func (ctrl *resizeController) markPVCAsFSResizeRequired(pvc *v1.PersistentVolume
 	return nil
 }
 
+// legacy markPVCResizeInProgress function, should be removed once RecoverFromVolumeExpansionFailure feature goes GA.
 func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
 	// Mark PVC as Resize Started
 	progressCondition := v1.PersistentVolumeClaimCondition{
@@ -520,7 +527,7 @@ func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeCl
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions,
 		[]v1.PersistentVolumeClaimCondition{progressCondition})
 
-	updatedPVC, err := ctrl.patchClaim(pvc, newPVC)
+	updatedPVC, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
 	if err != nil {
 		return nil, err
 	}
@@ -534,7 +541,7 @@ func (ctrl *resizeController) markPVCResizeFinished(
 	newPVC.Status.Capacity[v1.ResourceStorage] = newSize
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(pvc.Status.Conditions, []v1.PersistentVolumeClaimCondition{})
 
-	_, err := ctrl.patchClaim(pvc, newPVC)
+	_, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
 	if err != nil {
 		return fmt.Errorf("Mark PVC %q as resize finished failed: %v", util.PVCKey(pvc), err)
 	}
@@ -545,8 +552,8 @@ func (ctrl *resizeController) markPVCResizeFinished(
 	return nil
 }
 
-func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
-	patchBytes, err := util.GetPVCPatchData(oldPVC, newPVC)
+func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClaim, addResourceVersionCheck bool) (*v1.PersistentVolumeClaim, error) {
+	patchBytes, err := util.GetPVCPatchData(oldPVC, newPVC, addResourceVersionCheck)
 	if err != nil {
 		return nil, fmt.Errorf("can't patch status of PVC %s as generate path data failed: %v", util.PVCKey(oldPVC), err)
 	}
@@ -575,7 +582,11 @@ func (ctrl *resizeController) deletePreResizeCapAnnotation(pv *v1.PersistentVolu
 	return err
 }
 
-func (ctrl *resizeController) updatePVCapacity(pv *v1.PersistentVolume, oldCapacity, newCapacity resource.Quantity, fsResizeRequired bool) error {
+func (ctrl *resizeController) updatePVCapacity(
+	pv *v1.PersistentVolume,
+	oldCapacity, newCapacity resource.Quantity,
+	fsResizeRequired bool) (*v1.PersistentVolume, error) {
+
 	klog.V(4).Infof("Resize volume succeeded for volume %q, start to update PV's capacity", pv.Name)
 	newPV := pv.DeepCopy()
 	newPV.Spec.Capacity[v1.ResourceStorage] = newCapacity
@@ -590,11 +601,11 @@ func (ctrl *resizeController) updatePVCapacity(pv *v1.PersistentVolume, oldCapac
 		}
 	}
 
-	_, err := ctrl.patchPersistentVolume(pv, newPV)
+	updatedPV, err := ctrl.patchPersistentVolume(pv, newPV)
 	if err != nil {
-		return fmt.Errorf("updating capacity of PV %q to %s failed: %v", pv.Name, newCapacity.String(), err)
+		return pv, fmt.Errorf("updating capacity of PV %q to %s failed: %v", pv.Name, newCapacity.String(), err)
 	}
-	return nil
+	return updatedPV, nil
 }
 
 func (ctrl *resizeController) patchPersistentVolume(oldPV, newPV *v1.PersistentVolume) (*v1.PersistentVolume, error) {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -529,7 +529,7 @@ func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeCl
 
 	updatedPVC, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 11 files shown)


pkg/controller/expand_and_recover.go
```diff
@@ -0,0 +1,232 @@
...
+	if fsResizeRequired {
+		pvc, err = ctrl.markForPendingNodeExpansion(pvc)
+		return pvc, pv, err
+	}
+	pvc, err = ctrl.markOverallExpansionAsFinished(pvc, updatedSize)
+	return pvc, pv, err
+}
+
+// checks if pv can be expanded
+func (ctrl *resizeController) pvCanBeExpanded(pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
gnufied: I can do that, but personally I do not see much value in touching older functions. I would like to remove all of those functions once this feature goes beta in 1.24 (which is just few months away). So that is why, I have done side-by-side implementation rather than adding feature-gate checks all over the place, this way I can simply remove those functions when this feature goes beta.

pkg/controller/resize_status.go
```diff
@@ -0,0 +1,111 @@
...
+	updatedPVC, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
+	if err != nil {
+		return nil, err
+	}
+	return updatedPVC, nil
+}
+
+// markForPendingNodeExpansion is new set of functions designed around feature RecoverVolumeExpansionFailure
+// which correctly sets pvc.Status.ResizeStatus
+func (ctrl *resizeController) markForPendingNodeExpansion(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
gnufied: I would rather not jan, because I was thinking of deleting `markPVCAsFSResizeRequired` function altogether in next release.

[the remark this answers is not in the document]
gnufied: if I have to unify `markPVCAsFSResizeRequired` and `markForPendingNodeExpansion` then I will have to add feature-gate checks in `markPVCAsFSResizeRequired` - whereas having feature-gate check in `markForPendingNodeExpansion` is pointless. So the idea is - in next release I can delete `markPVCAsFSResizeRequired` entirely and `markForPendingNodeExpansion` remains the *only* way of marking a volume for pending node expansion.

pkg/controller/expand_and_recover.go
```diff
@@ -0,0 +1,232 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+)
+
+func (ctrl *resizeController) expandAndRecover(pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume) (*v1.PersistentVolumeClaim, *v1.PersistentVolume, error, bool) {
+	if !ctrl.pvCanBeExpanded(pv, pvc) {
+		klog.V(4).Infof("No need to resize PV %q", pv.Name)
+		return pvc, pv, nil, false
+	}
+	resizeNotCalled := false
```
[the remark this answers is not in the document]
gnufied: added a comment.

pkg/controller/expand_and_recover_test.go
```diff
@@ -0,0 +1,167 @@
...
+			name:                  "pvc.spec.size = pv.spec.size, recover_expansion=on",
+			pvc:                   getTestPVC("test-vol0", "1G", "1G", "", v1.PersistentVolumeClaimNoExpansionInProgress),
+			pv:                    createPV(1, "claim01", defaultNS, "test-uid", &fsVolumeMode),
+			recoverFeatureGate:    true,
+			expectedResizeStatus:  v1.PersistentVolumeClaimNodeExpansionPending,
+			expectedAllocatedSize: resource.MustParse("1G"),
+			expectResizeCall:      true,
+		},
+		{
+			name:                  "pvc.spec.size = pv.spec.size, recover_expansion=on",
```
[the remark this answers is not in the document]
gnufied: fixed.

pkg/controller/expand_and_recover_test.go
```diff
@@ -0,0 +1,167 @@
...
+		// expectations of test
+		expectedResizeStatus  v1.PersistentVolumeClaimResizeStatus
+		expectedAllocatedSize resource.Quantity
+		expectResizeCall      bool
+	}{
+		{
+			name:                  "pvc.spec.size > pv.spec.size, recover_expansion=on",
+			pvc:                   getTestPVC("test-vol0", "2G", "1G", "", v1.PersistentVolumeClaimNoExpansionInProgress),
+			pv:                    createPV(1, "claim01", defaultNS, "test-uid", &fsVolumeMode),
+			recoverFeatureGate:    true,
```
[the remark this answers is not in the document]
gnufied: removed

pkg/controller/expand_and_recover_test.go
```diff
@@ -0,0 +1,167 @@
...
+		},
+		{
+			name:                  "pvc.spec.size > pv.spec.size, recover_expansion=on, resize_status-node_expansion_inprogress",
+			pvc:                   getTestPVC("test-vol0", "10G", "1G", "3G", v1.PersistentVolumeClaimNodeExpansionInProgress),
+			pv:                    createPV(3, "claim01", defaultNS, "test-uid", &fsVolumeMode),
+			recoverFeatureGate:    true,
+			expectedResizeStatus:  v1.PersistentVolumeClaimNodeExpansionInProgress,
+			expectedAllocatedSize: resource.MustParse("3G"),
+			expectResizeCall:      false,
+		},
```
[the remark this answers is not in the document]
gnufied: Added two more unit tests that cover scenarios where value is actually shrunk.

pkg/controller/resize_status.go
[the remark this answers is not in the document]
jsafrane: ack

pkg/controller/expand_and_recover.go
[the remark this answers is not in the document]
jsafrane: I think it will need to be there until GA, the feature still can be disabled by a feature gate.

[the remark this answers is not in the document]
gnufied: ack - although existing implementation is beta too, but I think we can discuss this later.

pkg/controller/expand_and_recover_test.go
```diff
@@ -19,76 +19,86 @@ import (
...
+			expectedAllocatedSize: resource.MustParse("5G"),
+			expectResizeCall:      true,
+		},
+		{
+			name:                       "pvc.spec.size = pv.spec.size, resize_status=node_expansion_failed, disable_controller_expansion=true",
+			pvc:                        getTestPVC("test-vol0", "5G" /*specSize*/, "3G" /*statusSize*/, "10G" /*allocatedSize*/, v1.PersistentVolumeClaimNodeExpansionFailed),
+			pv:                         createPV(10, "claim01", defaultNS, "test-uid", &fsVolumeMode),
+			disableControllerExpansion: true,
+			expectedResizeStatus:       v1.PersistentVolumeClaimNodeExpansionPending,
+			expectedAllocatedSize:      resource.MustParse("5G"),
```
jsafrane, line 62: This is odd... how can be AllocatedSize shrunk after failed node expansion? PV size is 10Gi already.

jsafrane: Never mind, I see disableControllerExpansion. The test is correct.

jsafrane:
I tested it with AWS EBS (it allows one expansion of a volume in 6 hours, so it can throw nice errors). 
/lgtm
/hold
(for squash)

pkg/controller/expand_and_recover.go
```diff
@@ -0,0 +1,234 @@
...
+		return pvc, pv, err
+	}
+	pvc, err = ctrl.markOverallExpansionAsFinished(pvc, updatedSize)
+	return pvc, pv, err
+}
+
+// checks if pv can be expanded
+func (ctrl *resizeController) pvCanBeExpanded(pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) bool {
+	if !ctrl.resizer.CanSupport(pv, pvc) {
+		klog.V(4).Infof("Resizer %q doesn't support PV %q", ctrl.name, pv.Name)
```
jingxu97, line 201: maybe it can be a warning because you are calling resize, but resizer does not support it?

pkg/controller/controller.go
```diff
@@ -343,12 +343,18 @@ func (ctrl *resizeController) syncPVC(key string) error {
...
-		klog.V(4).Infof("No need to resize PV %q", pv.Name)
-		return nil
-	}
 
-	return ctrl.resizePVC(pvc, pv)
+	if utilfeature.DefaultFeatureGate.Enabled(features.RecoverVolumeExpansionFailure) {
+		_, _, err, _ := ctrl.expandAndRecover(pvc, pv)
+		return err
+	} else {
+		if !ctrl.pvNeedResize(pvc, pv) {
```
jingxu97, line 351: seem pvNeedResize can be called before checking feature is enabled?

jingxu97: is pvNeedResize more like pv cannot be resized due to condition is not met? Does it consider an error?

pkg/controller/controller.go
```diff
@@ -343,12 +343,18 @@ func (ctrl *resizeController) syncPVC(key string) error {
...
 		return nil
 	}
-	if !ctrl.pvNeedResize(pvc, pv) {
-		klog.V(4).Infof("No need to resize PV %q", pv.Name)
-		return nil
-	}
 
-	return ctrl.resizePVC(pvc, pv)
+	if utilfeature.DefaultFeatureGate.Enabled(features.RecoverVolumeExpansionFailure) {
+		_, _, err, _ := ctrl.expandAndRecover(pvc, pv)
```
jingxu97, line 348: why return pv, pvc, but without using use objects?

gnufied: for now, mainly for tests.

pkg/controller/expand_and_recover.go
gnufied: it can get called on all sort of PVs. such as intree pv,pvc etc, which are not resizable by external-resizer.  I think Info is suitable since most people will be running both intree expand_controller and this controller side-by-side for now.

pkg/controller/controller.go
gnufied: this is basically old flow - which checks if PV is resizable (such as, whether it is bound and bound to PVC which is being resized and whether it was already controller expanded and node expansion is pending). It is not a hard error if PV can not expanded at this moment. But anyways - this is old code and has kind always worked okay.

jingxu97: oh I confused about pvCanBeExpanded and pvNeedResize, thinking they are the same function.

pkg/controller/controller.go
```diff
@@ -520,9 +527,9 @@ func (ctrl *resizeController) markPVCResizeInProgress(pvc *v1.PersistentVolumeCl
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions,
 		[]v1.PersistentVolumeClaimCondition{progressCondition})
 
-	updatedPVC, err := ctrl.patchClaim(pvc, newPVC)
+	updatedPVC, err := ctrl.patchClaim(pvc, newPVC, true /* addResourceVersionCheck */)
 	if err != nil {
-		return nil, err
+		return updatedPVC, err
 	}
 	return updatedPVC, nil
```
xing-yang, line 534: Since `updatedPVC` is always returned now, we can remove the "if err != nil" code block and just have "return updatedPVC, err" as this will cover the case when err is nil as well.

pkg/controller/controller.go
```diff
@@ -545,19 +552,19 @@ func (ctrl *resizeController) markPVCResizeFinished(
 	return nil
 }
 
-func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
-	patchBytes, err := util.GetPVCPatchData(oldPVC, newPVC)
+func (ctrl *resizeController) patchClaim(oldPVC, newPVC *v1.PersistentVolumeClaim, addResourceVersionCheck bool) (*v1.PersistentVolumeClaim, error) {
```
jingxu97, line 555: could you add some comments about this option addResourceVersionCheck?

pkg/controller/controller.go
gnufied: Fixed.

pkg/controller/controller.go
gnufied: Fixed.

gnufied:
@jingxu97 @xing-yang addressed your comments. can you PTAL?

pkg/controller/resize_status.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return updatedPVC, nil
+}
+
+func (ctrl *resizeController) markControllerExpansionFailed(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
+	expansionFailedOnController := v1.PersistentVolumeClaimControllerExpansionFailed
+	newPVC := pvc.DeepCopy()
+	newPVC.Status.ResizeStatus = &expansionFailedOnController
+
+	updatedPVC, err := ctrl.patchClaim(pvc, newPVC, false /* addResourceVersionCheck */)
```
jingxu97, line 86: the reason no need to check resource version here?

gnufied: The general idea I had was if expansion failed we record the error even if we get a potential conflict. Since we are only patching  resize status we should be okay.

jingxu97: what is the result of conflict? It might overwrite some fields which are updated by other places?
how about we also check addResourceVersionCheck here? Any potential issue?

gnufied: The conflict when `addResourceVersionCheck` is present has nothing to do with content of the patch, but is about whether patch we are applying is with latest version of object or not. i.e - if two updates were performed on PVC concurrently and they updated different fields, then if `addResourceVersionCheck` is present with second patch - you will still get error (even if second update, changes entirely different field of PVC).  

> how about we also check addResourceVersionCheck here? Any potential issue?

We can, but if somehow our version of PVC was older then, the entire resizing operation has to be restarted before `ExpansionFailedOnController` can be set for `resizeStatus`. As such it does not affect the design but skipping `addResourceVersionCheck` is an optimization - so as we set this field if expansion fails on controller even if PVC was updated elsewhere.

gnufied: Added as a comment in the code.

PR #195, opened 2022-02-26
implement distributed resizing

### LINKED ISSUE #142: distributed resizing
For local storage (e.g. LVM disks) it makes sense to support a deployment model where the driver only runs locally on nodes together with sidecars. This is already support for provisioning (kubernetes-csi/external-provisioner#524, https://github.com/kubernetes-csi/external-provisioner#deployment-on-each-node).

travisghansen:
Signed-off-by: Travis Glenn Hansen <travisghansen@yahoo.com>

**What type of PR is this?**

/kind feature

Add resizing support to distributed setups to compliement provisioner and snapshotter features of the same nature.

1. Added a parameter "--node-deployment" as a command line option for resizer sidecar, which should be set to true when the sidecar is being deployed on a per node basis.
1. For these changes to work, NODE_NAME environment variable must also be set while deploying the sidecar controller.

Fixes #142 

```release-note
Adds support for distributed resizing.
```

### DIFF
--- a/cmd/csi-resizer/main.go
+++ b/cmd/csi-resizer/main.go
@@ -75,6 +75,8 @@ var (
 
 	handleVolumeInUseError = flag.Bool("handle-volume-inuse-error", true, "Flag to turn on/off capability to handle volume in use error in resizer controller. Defaults to true if not set.")
 
+	enableNodeDeployment = flag.Bool("node-deployment", false, "Enables deploying the sidecar controller together with a CSI driver on nodes to manage resizing for node-local volumes.")
+
 	featureGates map[string]bool
 
 	version = "unknown"
@@ -102,6 +104,19 @@ func main() {
 	}
 	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(featureGates); err != nil {
 		klog.Fatal(err)
+	}
+
+	// If distributed resizing is enabled and leaderElection is also set to true, return
+	if *enableNodeDeployment && *enableLeaderElection {
+		klog.Error("Leader election cannot happen when node-deployment is set to true")
+		os.Exit(1)
+	}
+
+	if *enableNodeDeployment {
+		node := os.Getenv("NODE_NAME")
+		if node == "" {
+			klog.Fatal("The NODE_NAME environment variable must be set when using --node-deployment.")
+		}
 	}
 
 	var config *rest.Config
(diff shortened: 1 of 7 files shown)


cmd/csi-resizer/main.go
```diff
@@ -156,7 +171,8 @@ func main() {
 		*timeout,
 		kubeClient,
 		informerFactory,
-		driverName)
+		driverName,
+		*enableNodeDeployment)
```
pohly, line 175: There is one potential scalability problem here: if there are now as many resizers running in the cluster as there are nodes, then the apiserver has to keep all of them informed about PV and PVC updates.

external-provisioner has the same problem, but it cannot really be avoided there because support for immediate provisioning depends on all provisioners seeing a new PVC.

I'm just wondering whether something better can be done for the other sidecars. The code below already checks for the `util.VolumeSelectedNodeKey` label. It would be possible to set up the PVC informer so that it does server-side filtering. It's doable, I just don't have a code example at hand.

That leaves the PV, though. external-provisioner would have to be modified to set a similar (the same?) label there.




Perhaps external-provisioner can set a special label for

pohly: Here is code which sets up an informer with server-side filtering:
https://github.com/kubernetes-csi/external-provisioner/blob/3b752c36ca71fbf5d7310bb6a568b93844062189/cmd/csi-provisioner/csi-provisioner.go#L452-L464

travisghansen: I thought about that as well but didn't want to 'kick against the pricks' as they say with how the others were implemented. I also noted that we are not currently doing server-side filtering on the driver either (which I found odd).

Given my limited `golang` skill set I would definitely need hand-holding on proper code to manage the watches/filters should we go that route.

mrpre:
Any update about this pr? I patched it and it seems the `resizer`  works well.

travisghansen:
I haven’t had a chance to address the scalability concerns yet unfortunately. It should work fine for small clusters for sure.

b8kings0ga:
wait to use this feature

silenceper:
@travisghansen Will this PR be merged? I also need this capability.

travisghansen:
I would like to see it be merged yes. When the PR was originally created there were concerns on the csi team of scalability issues with no real options for addressing them. I have not discussed this in a meeting with the csi team in a very long time but would be happy to discuss again and possibly get it over the hill.

Fundamentally the issue with scalability is about being able to limit the watches on each node to only volumes associated with that node. It's been a while since I looked but at the time this was originally put together there were no mechanisms to do so.

travisghansen:
I've done some research and 1 problem is that controller requires `capacity_range` and node does not. The csi test suite doesn't pass `capacity_range`, not sure what k8s normally does though.

PR #270, opened 2023-03-08
Update resize status

gnufied:
Update code to use new resize status

```release-note
Update recover from expansion failure to use newer APIs
```


pkg/controller/expand_and_recover.go
```diff
@@ -90,32 +88,31 @@ func (ctrl *resizeController) expandAndRecover(pvc *v1.PersistentVolumeClaim, pv
...
+		case v1.PersistentVolumeClaimControllerResizeInProgress,
+			v1.PersistentVolumeClaimControllerResizeFailed:
 			// This is case#2 or it could also be case#4 when user manually shrunk the PVC
 			// after expanding it.
 			if allocatedSize != nil {
 				newSize = *allocatedSize
 			}
 		default:
-			// It is impossible for ResizeStatus to be nil and allocatedSize to be not nil but somehow
+			// It is impossible for ResizeStatus to be empty and allocatedSize to be not nil but somehow
```
sunnylovestiramisu, line 112: Nit: the comment here is different from the other PR https://github.com/kubernetes/kubernetes/pull/116335. Should this be empty or ""?

gnufied: empty and "" are same thing right? the main thing is - corresponding intree code doesn't ever actually gets executed because all volume plugins that support volume expansion are actually now migrated to CSI.

pkg/controller/resize_status.go
```diff
@@ -36,12 +36,13 @@ func (ctrl *resizeController) markControllerResizeInProgress(
 		Status:             v1.ConditionTrue,
 		LastTransitionTime: metav1.Now(),
 	}
-	controllerExpansionInProgress := v1.PersistentVolumeClaimControllerExpansionInProgress
 	conditions := []v1.PersistentVolumeClaimCondition{progressCondition}
 
 	newPVC := pvc.DeepCopy()
 	newPVC.Status.Conditions = util.MergeResizeConditionsOfPVC(newPVC.Status.Conditions, conditions)
-	newPVC.Status.ResizeStatus = &controllerExpansionInProgress
+	newPVC.Status.AllocatedResourceStatuses = map[v1.ResourceName]v1.ClaimResourceStatus{
```
msau42, line 43: This overrides status for other resources if they exist. Ditto throughout

PR #317, opened 2023-08-08
Update dependency go modules for k8s v1.28.0-rc.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.28.0-rc.0


gnufied:
hmm, should we update this with https://github.com/kubernetes-csi/external-resizer/pull/270 ? I was doing that anyways as part of updating the sidecar

sunnylovestiramisu:
@gnufied I was not aware of you are working on it already, and yep if you are taking care if it in your PR#270, we do not need to do it in this PR.

gnufied:
Yeah I will have to modify some logic issues too while I am changing the code to use newer APIs. So it is best we handle this via #270

PR #332, opened 2023-09-07
Update dependency go modules for k8s v1.28.0

Sneha-at:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.28.0
            
```release-note
Update kubernetes dependencies to v1.28.0
```


go.mod
```diff
@@ -67,3 +67,17 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
```
Sneha-at, line 69: @msau42 Should I delete these replace statements here?

go.mod
```diff
@@ -6,17 +6,17 @@ require (
 	github.com/container-storage-interface/spec v1.8.0
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.12 // indirect
-	github.com/kubernetes-csi/csi-lib-utils v0.13.0
+	github.com/kubernetes-csi/csi-lib-utils v0.15.0
```
msau42, line 9: Keep this at 0.14.0 until we resolve the dependency issue.

go.mod
msau42: If the script is adding it then let's keep it. We should investigate how to modify the script to remove it if it's unnecessary.

go.mod
Sneha-at: This is causing dependency issue only for certain libraries from the otel package. It did not throw any error for "external-resizer" in `make test`. SHould we still use `v14` (may be to keep all sidecars on same dependency?) then we might want to as well revisit [node-driver-registrar](https://github.com/kubernetes-csi/node-driver-registrar/pull/323/files)

msau42: I think it's better to be consistent across all the sidecars.

Sneha-at: Updated `csi-lib-utils` to `v0.14.0`

PR #383, opened 2024-03-31
Allow CGO_ENABLED setting to be overwritten

mamccorm:
Allow GCO_ENABLED setting to be overridden, instead of hard-coding it as a CLI argument


jsafrane:
`release-tools` directory is imported from https://github.com/kubernetes-csi/csi-release-tools, please fix our makefiles in that project and then follow https://github.com/kubernetes-csi/csi-release-tools?tab=readme-ov-file#sharing-and-updating to update it here.

In addition, should it be `CGO_ENABLED?=0` , with a question mark?

Issue #367, opened 2024-01-26
1.30 - [KEP-3751] VolumeAttributesClass feature add tests for beta


sunnylovestiramisu:
There are a few different tasks:

1. Refactor the test files for external-resizer as comment in [PR](https://github.com/kubernetes-csi/external-resizer/pull/351/files)

2. Add more unit tests to cover error cases in [this discussion](https://kubernetes.slack.com/archives/C0615MXEYDU/p1706132168286789)

3. Add end to end functional tests by consuming csi-driver-host-path

AndrewSirenko:
Took a stab at item 1 in #402!

PR #351, opened 2023-11-10
Add Modify Volume Support

sunnylovestiramisu:
/kind feature

**What this PR does / why we need it**:
Implement logic of [this diagram](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/3751-volume-attributes-class/VolumeAttributesClass-ModifyVolume-Flow-v3.png)

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes-csi/external-resizer/issues/314

**Special notes for your reviewer**:

**Testing Steps**

***Positive Test Cases***
1. Using kubetest at with k8s 1.29 change including VAC
```
KUBE_FEATURE_GATES="VolumeAttributesClass=true" kubetest --up --runtime-config=api/all=true
```
2. Build the container image of [external-resizer with logs](https://github.com/kubernetes-csi/external-resizer/compare/master...sunnylovestiramisu:external-resizer:modifyVolumeTesting)
3. Set up csi-driver-host-path with the resizer image built in step 2 as in [this branch](https://github.com/sunnylovestiramisu/csi-driver-host-path/tree/testModifyVolume)
4. Follow [deploy-1.17-and-later.md runbook](https://github.com/sunnylovestiramisu/csi-driver-host-path/blob/master/docs/deploy-1.17-and-later.md) to deploy hostpath driver in the cluster
5. Create csi-pvc, csi-storageclass, csi-volumeattributesclass, and csi-app 
```
// csi-volumeattributesclass.yaml
apiVersion: storage.k8s.io/v1alpha1
kind: VolumeAttributesClass
metadata:
  name: silver
driverName: hostpath.csi.k8s.io
parameters:
  provisioned-iops: "3000"
```
6. Modify the pvc with ``kubectl edit pvc csi-pvc`` and add ``volumeAttributesClassName: silver`` to the pvc
7. Verify modify volume is successful

```
csi-driver-host-path git:(testModifyVolume) ✗ k describe pvc
Name:          csi-pvc
Namespace:     default
StorageClass:  csi-hostpath-sc
Status:        Bound
Volume:        pvc-561aca0d-a67b-4cb1-81c3-791b5c654423
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: hostpath.csi.k8s.io
               volume.kubernetes.io/storage-provisioner: hostpath.csi.k8s.io
[...]

PR #402, opened 2024-06-02
Refactor testutil MakePVC to implement builder pattern

AndrewSirenko:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR refactors test files for external-resizer as @carlory requested in [PR 351](https://github.com/kubernetes-csi/external-resizer/pull/351/files) so that VolumeAttributeClass feature can move to beta.

It makes `MakePVC` simpler by letting developers use use `MakePVC().WithXXX().WithYYY()` to implement an expected PVC.

Finally, this PR refactors helper functions previously in `testutil` to use this pattern and moves them to their respective packages as requested. 

Partially addresses #367

- Will squash commits before merging. 
- Unit tests pass locally. 
- Should I go refactor ALL unit tests to use this pattern? Or is this sufficient for marking this part of #367 off?

```release-note
NONE
```


carlory:
@AndrewSirenko Please refactor all the test code with the new library if possible. It makes the test code more readable and maintainable.

AndrewSirenko:
Hey @carlory I refactored most of the unit tests and also added in PVWrapper to `testutil` for consistency. I will do a final deep comb through for any places I've missed (like a few lines in `util_test.go`), but would you be able to skim through some of my changes to confirm I'm on the right track? 

Also, I've been keeping the convenience functions like `makeTestPVC` in each package intact. Is this fine or in an ideal world we'd want to get rid of them such that most `WithXXX()`s happen within the testcase definitions?

carlory:
Firstly, thanks for working on this task. It's not a blocker for the beta target.

The builder pattern wraps an object inside and returns the inner object after setting the properties via a series of chained methods. The following benefits are:

1. It can return a valid or invalid object. And 🌟🌟🌟  the object may be uncompleted, lacking some required properties but we don't take care of it in our test cases.
2. It is easy to generate the object. so we don't define lots of constructors 🌟🌟 for the object  
3. It provides a helper function to return a wrapper from an existing object. So we can generate a new object with a few properties changed from an existing object.

We can see lots of examples in `k8s.io/kubernetes`. i.e.

- https://github.com/kubernetes/kubernetes/blob/5bf1e95541d90e37f6c6637b5b45d8783e7907aa/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources_test.go#L103-L104
- https://github.com/kubernetes/kubernetes/blob/5bf1e95541d90e37f6c6637b5b45d8783e7907aa/pkg/scheduler/framework/plugins/nodename/node_name_test.go#L41

So I added a suggestion to the previous PR and hope to reduce the code duplication in the `pkg/modifycontroller` and `pkg/resizer` directories if possible.

Finally, after I deeply read the code and tried to add some changes, it may increase the complexity and the benefits are not obvious. So we can ignore this suggestion. What do you think? @AndrewSirenko 

cc @msau42

AndrewSirenko:
Closing PR, discussed that refactoring to builder pattern may be desirable once all sidecars are in one.

PR #418, opened 2024-07-15
Implement less noisy recovery

gnufied:
This is a WIP using - https://github.com/kubernetes/kubernetes/pull/126108


pkg/util/slowset.go
```diff
@@ -0,0 +1,94 @@
...
+func (s *SlowSet) Run(stopCh <-chan struct{}) {
+	for {
+		select {
+		case <-stopCh:
+			return
+		default:
+			time.Sleep(slowSetSleepDelay)
+			for key, t := range s.workSet {
+				if time.Since(t) > s.retentionTime {
+					s.Remove(key)
```
jsafrane, line 89: Volumes should be removed also after successful ControllerExpand or after non-infeasible (= feasible) error.

gnufied: Hmm, originally I wrote this code to slow down expansion retries for failing node-expansions too. So something like:

1. Node expansion fails with infeasible error.
2. Controller retries, but adds the volume to `slowSet`.
3. Controller expansion succeeds and removes the volume from `slowSet`(*desired*)
4. Node expansion fails again with infeasible error.
5. Controller will immediately retry the expansion (*undesired*)

I agree though not removing volume from `slowSet` has a downside:

1. Controller expansion fails with infeasible error.
2. Controller retries and adds the volume to `slowSet`
3. Controller expansion somehow succeeds, but saving of one or more objects fail (because of resourceVersion check or something)
4. Now - controller will not correct this error until next 5 minutes (*undesired*)

A way to solve this would be perhaps to use - different sets for node and controller failures.

go.mod
```diff
@@ -52,38 +52,30 @@ require (
...
-
-replace k8s.io/apiserver => k8s.io/apiserver v0.30.0
-
-replace k8s.io/client-go => k8s.io/client-go v0.30.0
-
-replace k8s.io/component-base => k8s.io/component-base v0.30.0
-
 replace k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.30.0
 
 replace k8s.io/kms => k8s.io/kms v0.30.0
```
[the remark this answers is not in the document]
gnufied: fixed. removed `kms` dependency. Not sure how it crept in.

go.mod
```diff
@@ -3,30 +3,30 @@ module github.com/kubernetes-csi/external-resizer
...
+	github.com/kubernetes-csi/csi-lib-utils v0.19.0-beta.0
 	golang.org/x/oauth2 v0.21.0 // indirect
-	golang.org/x/term v0.22.0 // indirect
+	golang.org/x/term v0.21.0 // indirect
 	google.golang.org/grpc v1.65.0
-	k8s.io/api v0.30.0
-	k8s.io/apimachinery v0.30.0
+	k8s.io/api v0.31.0-rc.0
+	k8s.io/apimachinery v0.31.0-rc.0
 	k8s.io/apiserver v0.30.0
```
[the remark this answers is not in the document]
gnufied: bumped everything to v1.31

pkg/util/slowset.go
gnufied: Even if we solved this for controller expansion operations, it is much trickier to solve this for node expansion failures.

gnufied: okay, so I implemented removal via - https://github.com/kubernetes-csi/external-resizer/pull/418/commits/88252668b58ee62b35d8f3c1de75bb75774afecd

pkg/util/slowset.go
```diff
@@ -25,68 +25,86 @@ const (
...
+	Timestamp   time.Time
+}
+
+type SlowSet[T ~string] struct {
 	sync.RWMutex
 	// retentionTime is the time after which an item will be removed from the set
 	// this indicates, how long before an operation on pvc can be retried.
 	retentionTime time.Duration
-	workSet       map[string]time.Time
+	workSet       map[string]FailureTimestamp[T]
```
jsafrane, line 38: Why the slowset has multiple entries per PVC?
Either a PVC is in the slowset (the last expansion was infeasible) or it's not (anything else - success, feasible error, resize has not started yet).

pkg/controller/expand_and_recover.go
```diff
@@ -191,7 +191,7 @@ func (ctrl *resizeController) expandAndRecover(pvc *v1.PersistentVolumeClaim, pv
 func (ctrl *resizeController) markForSlowRetry(pvcKey string, resizeStatus v1.ClaimResourceStatus) {
 	if resizeStatus == v1.PersistentVolumeClaimControllerResizeInfeasible ||
 		resizeStatus == v1.PersistentVolumeClaimNodeResizeInfeasible {
-		ctrl.slowSet.Add(pvcKey)
+		ctrl.slowSet.Add(pvcKey, resizeStatus)
```
jsafrane, line 194: This is the only Add call and it's always called with `PersistentVolumeClaimNodeResizeInfeasible`, so why all the code around multiple states?

gnufied: No it can be called with either `NodeResizeInfeasible` or `ControllerResizeInfeasible`. The idea is to not accidentally clear `NodeResizeInfeasible` errors if controller expansion is successful or fails with a non-infeasible error.

pkg/util/slowset.go
```diff
@@ -25,68 +25,86 @@ const (
 	slowSetSleepDelay = 100 * time.Millisecond
 )
 
-type SlowSet struct {
+type FailureTimestamp[T ~string] struct {
+	FailureType T
+	Timestamp   time.Time
+}
+
+type SlowSet[T ~string] struct {
```
jsafrane, line 33:
```
// SlowSet is a set of API objects that should be synced at slower rate. Key is typically the object
// namespace + name. 
// Reason why is the object in the queue is recorded in type `T`. For example,
// `SlowSet[v1.ClaimResourceStatus]` will remember if the object is in the queue because it status is
// `PersistentVolumeClaimControllerResizeInfeasible` or `PersistentVolumeClaimNodeResizeInfeasible`.
// `Remove()` will then be able to remove the object from the slow set when ControllerResize becomes
// feasible or NodeResize becomes feasible.
```

gnufied:
@jsafrane okay I have pushed a change that completely removes attempts to slow retries of recovery from `NodeInfeasible` volume errors. We will revisit this aspect later.

go.mod
```diff
@@ -6,27 +6,27 @@ require (
 	github.com/container-storage-interface/spec v1.10.0
 	github.com/google/go-cmp v0.6.0
 	github.com/google/gofuzz v1.2.0 // indirect
-	github.com/imdario/mergo v0.3.12 // indirect
-	github.com/kubernetes-csi/csi-lib-utils v0.18.1
-	golang.org/x/oauth2 v0.21.0 // indirect
-	golang.org/x/term v0.22.0 // indirect
+	github.com/imdario/mergo v0.3.16 // indirect
+	github.com/kubernetes-csi/csi-lib-utils v0.19.0-beta.0
```
jsafrane, line 10: this should be 0.19.0 now

pkg/util/slowset.go
```diff
@@ -0,0 +1,96 @@
...
+	return 0
+}
+
+func (s *SlowSet) Run(stopCh <-chan struct{}) {
+	for {
+		select {
+		case <-stopCh:
+			return
+		default:
+			time.Sleep(slowSetSleepDelay)
```
jsafrane, line 88: This does not work well with `TimeRemaining()`.

When a PVC enters the slow set, its resyns is scheduled after `TimeRemaining()`. In the first sync after `TimeRemainig()`, `Contains()` returns true, because the PVC is still in the set, but it's `TimeRemaining()` is zero, so the PVC gets resynced immediately. And again and again, until this `Sleep` finally wakes up an removes expired PVCs from the slow queue. I've seen hundreds of useless syncs.

One of the possible fixes would be to check remaining time in `Contains()` and remove expired items from the set there. Of remove the more often here.

go.mod
gnufied: fixed.

pkg/util/slowset.go
gnufied: I added an additional check in `Contains` function.

huww98:
This PR bumps go version to 1.22.5, which breaks our internal CI. Should we also update this line? We use this to determine the go version used to build it.

https://github.com/kubernetes-csi/external-resizer/blob/dc9173e30380cf4b68e7729a562d22ab2d210bbe/release-tools/prow.sh#L89

jsafrane:
I'l updating release-tools in https://github.com/kubernetes-csi/external-resizer/pull/431

PR #419, opened 2024-07-22
Skip modifying volumes owned by other CSI drivers

ConnorJC3:
**What type of PR is this?**

/kind bug

Fixes issue (3) from https://github.com/kubernetes/kubernetes/pull/124151#issuecomment-2152647568 - Today, the `external-resizer` will attempt to modify _any_ PVC awaiting modification without checking for if it matches the local CSI driver.

N/A

No test exercises `syncPVC`, so I could not update the tests for this case. I have manually tested the change against multiple drivers.

```release-note
Fixed a bug causing external-resizer to attempt to modify volumes owned by other CSI drivers
```

### DIFF
--- a/pkg/modifycontroller/controller.go
+++ b/pkg/modifycontroller/controller.go
@@ -277,8 +277,9 @@ func (ctrl *modifyController) syncPVC(key string) error {
 	// Only trigger modify volume if the following conditions are met
 	// 1. Non empty vac name
 	// 2. PVC is in Bound state
+	// 3. PV CSI driver name matches local driver
 	vacName := pvc.Spec.VolumeAttributesClassName
-	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound {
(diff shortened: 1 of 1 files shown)


pkg/modifycontroller/controller.go
```diff
@@ -277,8 +277,9 @@ func (ctrl *modifyController) syncPVC(key string) error {
 	// Only trigger modify volume if the following conditions are met
 	// 1. Non empty vac name
 	// 2. PVC is in Bound state
+	// 3. PV CSI driver name matches local driver
 	vacName := pvc.Spec.VolumeAttributesClassName
-	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound {
+	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound && pv.Spec.CSI.Driver == ctrl.name {
```
msau42, line 282: can we add a unit test for `syncPVC` if one doesn't exist?

PR #420, opened 2024-07-24
Add extra-modify-metadata flag to external-resizer sidecar

mdzraf:
**What type of PR is this?**
/kind feature

This PR adds a flag that can be set by the user which will inject extra PVC and PV metadata as parameters when calling ModifyVolume on CSI Drivers. This will enable CSI Drivers to support StorageClass parameters that depend on the extra metadata in modification via VolumeAttributesClass as well. 

See [doc](https://docs.google.com/document/d/1SI8RDwNmzyaWtgaoglqQ3ydfye_6Vwy_EF1rVdaHQb8/edit?usp=sharing) for more info

N/A

N/A

```release-note
Adds the flag --extra-modify-metadata, which when set to true, will inject extra PVC and PV metadata as parameters when calling ModifyVolume on CSI Drivers
```

### DIFF
--- a/pkg/modifycontroller/modify_status_test.go
+++ b/pkg/modifycontroller/modify_status_test.go
@@ -102,7 +102,7 @@ func TestMarkControllerModifyVolumeStatus(t *testing.T) {
 		tc := test
 		t.Run(tc.name, func(t *testing.T) {
 			featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.VolumeAttributesClass, true)
-			client := csi.NewMockClient("foo", true, true, true, true, true)
+			client := csi.NewMockClient("foo", true, true, true, true, true, false)
 			driverName, _ := client.GetDriverName(context.TODO())
 
 			pvc := test.pvc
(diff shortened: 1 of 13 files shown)


gnufied:
Can you rebase this PR?

mdzraf:
@gnufied I have rebased the PR and all tests have passed, it is ready for review

pkg/modifycontroller/modify_status_test.go
```diff
@@ -162,7 +162,7 @@ func TestUpdateConditionBasedOnError(t *testing.T) {
 		tc := test
 		t.Run(tc.name, func(t *testing.T) {
 			featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.VolumeAttributesClass, true)
-			client := csi.NewMockClient("foo", true, true, true, true, true)
+			client := csi.NewMockClient("foo", true, true, true, true, true, false)
```
gnufied, line 165: Are we even using this boolean value anywhere?

gnufied: Also another nit (and not this PRs fault), I am beginning to dislike the endless list of true/false arguments with no idea, what they do. :(

mdzraf: That boolean value is not used in most of the tests but I had to make the necessary changes to the MockClient in order to be able to write unit tests for the flag on the `TestModify`, I did not think it was appropriate to make a new struct to deal with the `TestModify` tests only, but I could.

I do agree with you on that it is messy to have so many unlabeled booleans, do you have any suggestions?

mdzraf: @gnufied Are there any other outstanding concerns with the PR besides this?

PR #435, opened 2024-08-21
Add changelog for v1.12

gnufied:
Add changelog for v1.12


gnufied:
hmm. yeah I was thinking about it. I think we can update it.

PR #459, opened 2024-12-20
Move recover beta new

gnufied:
xref #1790

```release-note
Move RecoverVolumeExpansionFailure to beta and enabled by default.
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,4 +1,4 @@
 #! /bin/bash
-
+export CSI_PROW_E2E_VERSION="release-1.31"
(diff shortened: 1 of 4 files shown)


.prow.sh
```diff
@@ -1,4 +1,4 @@
 #! /bin/bash
-
+export CSI_PROW_E2E_VERSION="release-1.31"
```
jsafrane, line 2: Can you please comment why is this necessary and when we should remove it?

gnufied: added

PR #471, opened 2025-02-13
Fix modifycontroller panic for non-csi volumes

### LINKED ISSUE #470: Cannot modify csi-migrated volumes (leads to nil-pointer panic)
x-ref: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2340 (credit to @gpapakyriakopoulos) 

## Problem 

PVCs that were provisioned by an in-tree driver cannot be modified via VAC. Attempting to do so will throw external-resizer into crashBackoffLoop,. 

The [modifycontroller code](https://github.com/kubernetes-csi/external-resizer/blob/20072c0fdf8baaf919ef95d6e918538ba9d84eaf/pkg/modifycontroller/controller.go#L275) assumes that the PV resource associated with the modified volume has field `PV.Spec.CSI.Driver`. But PV resources associated with volumes that were migrated to csi drivers csi-migration volumes do not have this field, leading to a [nil pointer exception](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2340#issuecomment-2657496553). 

Preventing a panic with an additional nil check is trivial (which I can provide a PR for shortly), but we should discuss what to do in the case of migrated volumes (do we want to fallback to checking `pv.kubernetes.io/migrated-to` annotation?).

In the meantime I have provided a risky workaround in my comment below. 

## Relevant snippets

```
func (ctrl *modifyController) syncPVC(key string) error {
...
	// Only trigger modify volume if the following conditions are met
	// 1. Non-empty vac name
	// 2. PVC is in Bound state
	// 3. PV CSI driver name matches local driver
	vacName := pvc.Spec.VolumeAttributesClassName
	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound && pv.Spec.CSI.Driver == ctrl.name {
		_, _, err, _ := ctrl.modify(pvc, pv)
		if err != nil {
			return err
		}
}
```

```
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/migrated-to: ebs.csi.aws.com
    pv.kubernetes.io/provisioned-by: kubernetes.io/aws-ebs
...
...
spec:
  awsElasticBlockStore:
    fsType: ext4
    volumeID: vol-045ec457863e5980b
...
```

## Reproduction steps 

Can reproduce with EBS CSI Driver v1.39.0 on EKS 1.31:

```
> kubectl apply -f - <<EOF
apiVersion: storage.k8s.io/v1beta1
kind: VolumeAttributesClass
metadata:
  name: gp3-class
driverName: ebs.csi.aws.com
parameters:
  type: gp3
  throughput: "128"
  iops: "3000"
  encrypted: "true"
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
volumeBindingMode: WaitForFirstConsumer
parameters:
  type: gp2
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim1
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 10Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: app
    image: centos
    command: ["/bin/sh"]
    args: ["-c", "while true; do echo $(date -u) | tee /data/out.txt; sleep 5; done"]
    volumeMounts:
    - name: persistent-storage
      mountPath: /data
  volumes:
  - name: persistent-storage
    persistentVolumeClaim:
      claimName: ebs-claim1
EOF

> kubectl patch pvc ebs-claim1 --patch '{"spec": {"volumeAttributesClassName": "gp3-class"}}
```

AndrewSirenko:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind bug

PVCs that were provisioned by an in-tree driver cannot be modified via VAC. Attempting to do so will throw external-resizer into crashBackoffLoop,.

The [modifycontroller code](https://github.com/kubernetes-csi/external-resizer/blob/20072c0fdf8baaf919ef95d6e918538ba9d84eaf/pkg/modifycontroller/controller.go#L275) assumes that the PV resource associated with the modified volume has field PV.Spec.CSI.Driver. But PV resources associated with volumes that were migrated to csi drivers csi-migration volumes do not have this field, leading to panic and controller crashBackoff loop.

Preventing a panic with an additional nil check is trivial. Also added a simple unit test for this case. 

Fixes #470 

Tested with custom built image that nothing happens when you patch a non-csi volume.

```release-note
Fix panic when attempting to modify non-CSI volumes
```

### DIFF
--- a/pkg/modifycontroller/controller.go
+++ b/pkg/modifycontroller/controller.go
@@ -268,11 +268,16 @@ func (ctrl *modifyController) syncPVC(key string) error {
 	}
 
 	// Only trigger modify volume if the following conditions are met
-	// 1. Non-empty vac name
-	// 2. PVC is in Bound state
-	// 3. PV CSI driver name matches local driver
+	// 1. PV provisioned by CSI driver AND driver name matches local driver
+	// 2. Non-empty vac name
+	// 3. PVC is in Bound state
+	if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != ctrl.name {
+		klog.V(7).InfoS("Skipping PV provisioned by different driver", "PV", klog.KObj(pv))
+		return nil
+	}
+
 	vacName := pvc.Spec.VolumeAttributesClassName
-	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound && pv.Spec.CSI.Driver == ctrl.name {
+	if vacName != nil && *vacName != "" && pvc.Status.Phase == v1.ClaimBound {
 		_, _, err, _ := ctrl.modify(pvc, pv)
 		if err != nil {
 			return err
--- a/pkg/modifycontroller/controller_test.go
+++ b/pkg/modifycontroller/controller_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kubernetes-csi/external-resizer/pkg/util"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 	"testing"
 	"time"
@@ -156,6 +157,25 @@ func TestSyncPVC(t *testing.T) {
 	pvcWithUncreatedPV := createTestPVC(pvcName, targetVac /*vacName*/, testVac /*curVacName*/, testVac /*targetVacName*/)
 	pvcWithUncreatedPV.Spec.VolumeName = ""
 
+	nonCSIPVC := &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{Name: pvcName, Namespace: pvcNamespace},
+		Spec: v1.PersistentVolumeClaimSpec{
+			VolumeAttributesClassName: &targetVac,
+			VolumeName:                pvName,
+		},
+		Status: v1.PersistentVolumeClaimStatus{
+			Phase: v1.ClaimBound,
+		},
+	}
+	nonCSIPV := &v1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pvName,
+		},
+		Spec: v1.PersistentVolumeSpec{
+			VolumeAttributesClassName: nil,
+		},
+	}
+
 	tests := []struct {
 		name          string
 		pvc           *v1.PersistentVolumeClaim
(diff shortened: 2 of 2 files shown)


pkg/modifycontroller/controller_test.go
```diff
@@ -156,6 +157,25 @@ func TestSyncPVC(t *testing.T) {
 	pvcWithUncreatedPV := createTestPVC(pvcName, targetVac /*vacName*/, testVac /*curVacName*/, testVac /*targetVacName*/)
 	pvcWithUncreatedPV.Spec.VolumeName = ""
 
+	nonCSIPVC := &v1.PersistentVolumeClaim{
```
AndrewSirenko, line 160: This felt cleaner than creating yet another createTestPVCForNonMigratedVolumes function, but open to suggestions.

pkg/modifycontroller/controller.go
```diff
@@ -268,11 +268,16 @@ func (ctrl *modifyController) syncPVC(key string) error {
...
 
 	// Only trigger modify volume if the following conditions are met
-	// 1. Non-empty vac name
-	// 2. PVC is in Bound state
-	// 3. PV CSI driver name matches local driver
+	// 1. PV provisioned by CSI driver AND driver name matches local driver
+	// 2. Non-empty vac name
+	// 3. PVC is in Bound state
+	if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != ctrl.name {
+		klog.V(7).InfoS("Skipping PV provisioned by different driver", "PV", klog.KObj(pv))
```
sunnylovestiramisu, line 275: Overall LGTM. Question, why this is level 7 verbosity?

AndrewSirenko: We don't want drivers to log each time their resizer skips processing a PV provisioned by a different driver. It is too noisy, even at verbosity 4. 

However this information might be useful in a troubleshooting setting (Seeing which drivers are skipping which PVs).

AndrewSirenko:
.cherry-pick release-1.12

PR #462, opened 2025-01-08
Upgrade dependencies ahead of external-resizer v1.13.0

AndrewSirenko:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

Upgrade dependencies ahead of external-resizer v1.13.0

Note that new version of csi-translation-lib supports contextual logging, but resizer does not. Therefore I passed down a klog.TODO() for now. 

Supercedes #452, #451 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #490, opened 2025-03-26
update client go to the latest version

### LINKED ISSUE #489: client-go version set to a really old version (v1.5.2) in go mod
The client-go version that is set in go mod seems to be 4 years old 
https://github.com/kubernetes/client-go/releases/tag/v1.5.2

This seems to have been updated as part of 

https://github.com/kubernetes-csi/external-resizer/pull/462

Is this expected or was that a mistake ?  This creates issue when we try to import this library in our go projects, since this library has a very old client-go as a dependency

kvalliyurnatt:
**What type of PR is this?**

/kind bug

This PR updates the client-go version to v0.32.3, currently this is set to v1.5.2 which is a really old version of client-go. Not sure if this was [introduced](https://github.com/kubernetes-csi/external-resizer/pull/462) due to some specific need. 

Fixes #489 

NONE


sunnylovestiramisu:
Interesting why dependabot does not update it?

/lgtm

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	k8s.io/api v0.32.2
 	k8s.io/apimachinery v0.32.2
 	k8s.io/apiserver v0.32.0
-	k8s.io/client-go v1.5.2
+	k8s.io/client-go v0.32.3
```
msau42, line 16: The client-go version is overridden by L84, so this change theoretically shouldn't do anything

mowangdk: > The client-go version is overridden by L84, so this change theoretically shouldn't do anything

I think it's better to remove replace in go.mod if not necessary? Actually, I found this issue on CSI AIO testing. And this could affect users relying on this repository.

kvalliyurnatt: I removed the replace on L84

huww98: I think we should just remove all the replace statements. We only need them if we import `k8s.io/kubernetes` (which is not supported). But we have never imported that from git history of go.mod.

go.mod
```diff
@@ -10,10 +10,10 @@ require (
 	golang.org/x/oauth2 v0.25.0 // indirect
 	golang.org/x/term v0.28.0 // indirect
```
huww98, line 11: I think it is strange that these 3 indirect requires are not in the block below. Maybe we can take the chance to fix this too.

go.mod
```diff
@@ -81,8 +81,6 @@ replace k8s.io/apimachinery => k8s.io/apimachinery v0.32.0
 
 replace k8s.io/apiserver => k8s.io/apiserver v0.32.0
 
-replace k8s.io/client-go => k8s.io/client-go v0.32.0
-
 replace k8s.io/component-base => k8s.io/component-base v0.32.0
```
sunnylovestiramisu, line 84: I think all these replace can be removed?

go.mod
mowangdk: It's doesn't match the issue, We can fix it individually

go.mod
mowangdk: Agreed

sunnylovestiramisu: Then why don't we just remove these?

PR #494, opened 2025-04-17
Removing Slowset Utility

mdzraf:
**What type of PR is this?**

/kind cleanup

This PR removes the slowset utility out of the external-resizer into csi-lib-utils so that other sidecars can also use this utility. 

~This PR must not be merged until this [csi-lib-utils PR](https://github.com/kubernetes-csi/csi-lib-utils/pull/192) is merged. Currently in this code, it is pointing to a local version of csi-lib-utils that was being used for testing, this will change to point to the actual csi-lib-utils once the slowset is merged into csi-lib-utils.~

```release-note
Removed the slowset utility and moved it to the shared csi-lib-utils.
```


gnufied:
I am in middle of making 1.14 release. Lets finish that and then we can merge this.

PR #513, opened 2025-07-20
modify: support rollback

huww98:
**What type of PR is this?**

/kind feature

>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Support rollback to VAC A if modifying from A to B failed with a final error.
This works just like we modifying it again to C on final error.

The significant changes in the sync logic:
- Always retry if pvc.Status.ModifyVolumeStatus is not nil, which means the
  last transation does not finish successfully.
- Keep reconciling to the previous target if spec is rolled back to nil, until
  it succeeds or we get an infeasible error. Then we just leave it at its
  current state and stop reconciling for it, since user may not care about it
  now.

See https://github.com/kubernetes/enhancements/pull/5482 for details

Fixes #

The first 2 commits are already included in my other PRs. Please review the last commit. I will rebase when the previous PRs are merged.

```release-note
Support rolling back when modify failed
```

### DIFF
--- a/pkg/modifycontroller/controller_test.go
+++ b/pkg/modifycontroller/controller_test.go
@@ -204,6 +204,11 @@ func TestSyncPVC(t *testing.T) {
 			callCSIModify: true,
 		},
 		{
+			name:          "Should NOT modify if PVC deleted",
+			pvc:           nil,
+			callCSIModify: false,
+		},
+		{
 			name:          "Should NOT modify if PVC not in bound state",
 			pvc:           unboundPVC,
 			pv:            basePV,
@@ -227,7 +232,13 @@ func TestSyncPVC(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			client := csi.NewMockClient(testDriverName, true, true, true, true, true)
 
-			initialObjects := []runtime.Object{test.pvc, test.pv, testVacObject, targetVacObject}
+			initialObjects := []runtime.Object{testVacObject, targetVacObject}
+			if test.pvc != nil {
+				initialObjects = append(initialObjects, test.pvc)
+			}
+			if test.pv != nil {
+				initialObjects = append(initialObjects, test.pv)
+			}
 			ctrlInstance := setupFakeK8sEnvironment(t, client, initialObjects)
 
 			err := ctrlInstance.syncPVC(pvcNamespace + "/" + pvcName)
--- a/pkg/modifycontroller/controller_test.go
+++ b/pkg/modifycontroller/controller_test.go
@@ -23,13 +23,13 @@ import (
 )
 
 func TestController(t *testing.T) {
-	basePVC := createTestPVC(pvcName, testVac /*vacName*/, testVac /*curVacName*/, testVac /*targetVacName*/)
+	basePVC := createTestPVC(pvcName, testVac /*vacName*/, testVac /*curVacName*/, "" /*targetVacName*/)
+	basePVC.Status.ModifyVolumeStatus = nil
 	basePV := createTestPV(1, pvcName, pvcNamespace, "foobaz" /*pvcUID*/, &fsVolumeMode, testVac)
 	firstTimePV := basePV.DeepCopy()
 	firstTimePV.Spec.VolumeAttributesClassName = nil
 	firstTimePVC := basePVC.DeepCopy()
 	firstTimePVC.Status.CurrentVolumeAttributesClassName = nil
-	firstTimePVC.Status.ModifyVolumeStatus = nil
 
 	tests := []struct {
 		name          string
@@ -151,6 +151,9 @@ func TestSyncPVC(t *testing.T) {
 	pvcWithUncreatedPV := createTestPVC(pvcName, targetVac /*vacName*/, testVac /*curVacName*/, testVac /*targetVacName*/)
 	pvcWithUncreatedPV.Spec.VolumeName = ""
 
+	inprogressPVC := createTestPVC(pvcName, "" /*vacName*/, "" /*curVacName*/, testVac /*targetVacName*/)
+	inprogressPVC.Status.ModifyVolumeStatus.Status = v1.PersistentVolumeClaimModifyVolumeInProgress
+
 	nonCSIPVC := &v1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{Name: pvcName, Namespace: pvcNamespace},
 		Spec: v1.PersistentVolumeClaimSpec{
@@ -183,16 +186,22 @@ func TestSyncPVC(t *testing.T) {
 			callCSIModify: true,
 		},
 		{
+			name:          "Should NOT modify when rollback to empty VACName",
+			pvc:           createTestPVC(pvcName, "" /*vacName*/, "" /*curVacName*/, testVac /*targetVacName*/),
+			pv:            basePV,
+			callCSIModify: false,
+		},
+		{
 			name:          "Should NOT modify if PVC managed by another CSI Driver",
 			pvc:           basePVC,
 			pv:            otherDriverPV,
 			callCSIModify: false,
 		},
 		{
-			name:          "Should NOT modify if PVC has empty Spec.VACName",
-			pvc:           createTestPVC(pvcName, "" /*vacName*/, testVac /*curVacName*/, testVac /*targetVacName*/),
+			name:          "Should execute ModifyVolume for InProgress target if PVC has empty Spec.VACName",
+			pvc:           inprogressPVC,
 			pv:            basePV,
-			callCSIModify: false,
+			callCSIModify: true,
 		},
 		{
 			name:          "Should NOT modify if PVC not in bound state",
--- a/pkg/modifycontroller/controller_test.go
+++ b/pkg/modifycontroller/controller_test.go
@@ -1,9 +1,9 @@
 package modifycontroller
 
 import (
-	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
@@ -14,7 +14,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
-	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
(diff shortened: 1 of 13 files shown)


pkg/modifycontroller/modify_volume.go
```diff
@@ -52,59 +52,77 @@ func (ctrl *modifyController) modify(pvc *v1.PersistentVolumeClaim, pv *v1.Persi
...
+		klog.V(4).InfoS("stop reconcile for rolled back PVC", "PV", klog.KObj(pv))
+		// Don't try to revert anything here, because we only record the result of the last modification.
+		// We don't know what happened before. User can switch between InProgress/Infeasible/Pending status
+		// freely by modifying the spec.
+		return pvc, pv, nil, false
+	}
+
+	// Check if we should change our target
+	_, inUncertainState := ctrl.uncertainPVCs.Load(pvcKey)
+	if (status != nil && status.Status == v1.PersistentVolumeClaimModifyVolumeInProgress && inUncertainState) || pvcSpecVacName == "" {
```
sunnylovestiramisu, line 77: If status.Status == v1.PersistentVolumeClaimModifyVolumeInProgress alone should be a sufficient condition to not modify volume anymore?

gnufied: that condition exists to ensure that for transient failing modifications we keep trying to reconcile to whatever value was recorded in `target` and do not start with new value of `vac`.

pkg/modifycontroller/modify_volume.go
```diff
@@ -53,59 +52,77 @@ func (ctrl *modifyController) modify(pvc *v1.PersistentVolumeClaim, pv *v1.Persi
...
```
gnufied, line 77: So I think there is a lot happening in this conditional, but some comments:

1. `status` has to be not nil for not `pvcSpec == ""` case and uncertain case. So may be, pull the `status != nil` check above.
2. Same as my comment in other conditional. I personally find these complex conditions hard to read in future unless I am really paying attention and having these complex conditions behind a function that returns boolean value makes this greatly more readable.
3.

pkg/modifycontroller/modify_volume.go
```diff
@@ -53,59 +52,77 @@ func (ctrl *modifyController) modify(pvc *v1.PersistentVolumeClaim, pv *v1.Persi
...
+
+	// Mark pvc.Status.ModifyVolumeStatus as in progress
+	pvc, err = ctrl.markControllerModifyVolumeStatus(pvc, v1.PersistentVolumeClaimModifyVolumeInProgress, nil)
+	if err != nil {
+		return pvc, pv, err, false
+	}
+	// Record an event to indicate that external resizer is modifying this volume.
+	ctrl.eventRecorder.Event(pvc, v1.EventTypeNormal, util.VolumeModify,
+		fmt.Sprintf("external resizer is modifying volume %s with vac %s", pvc.Name, vac.Name))
+	return ctrl.controllerModifyVolumeWithTarget(pvc, pv, vac)
```
gnufied, line 125: Can you fix the implementation in `controllerModifyVolumeWithTarget`, so as we aren't updating PVC twice for infeasible errors? (basically same as comment I left on 512).

pkg/modifycontroller/controller_test.go
```diff
@@ -321,6 +328,65 @@ func TestInfeasibleRetry(t *testing.T) {
...
+		waitCount int
+		err       error
+	}{
+		// TODO: This case is flaky due to fake client lacks resourceVersion support.
+		// {
+		// 	name:      "success",
+		// 	waitCount: 10,
+		// },
+		{
+			name:      "uncertain",
```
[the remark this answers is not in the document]
huww98: I want to give a note to others that this will not work now. To save them some time if they want to add a success case. A test that only include failure case but not success case is uncommon. Technically we may make this work by adding a custom reactor to the fake client to check resourceVersion.

pkg/modifycontroller/controller_test.go
```diff
@@ -321,6 +328,65 @@ func TestInfeasibleRetry(t *testing.T) {
...
+									Driver:       testDriverName,
+									VolumeHandle: fmt.Sprintf("foo-%d", i),
+								},
+							},
+						},
+					},
+				)
+			}
+			ctrlInstance := setupFakeK8sEnvironment(t, client, initialObjects)
+			go ctrlInstance.Run(3, t.Context(), nil)
```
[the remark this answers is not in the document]
huww98: In fact, I'm not sure why we need to expose the wg here. Should this just work?
```diff
 func (ctrl *modifyController) Run(
-       workers int, ctx context.Context, wg *sync.WaitGroup) {
+       workers int, ctx context.Context) {
        defer ctrl.claimQueue.ShutDown()
 
        klog.InfoS("Starting external resizer for modify volume", "controller", ctrl.name)
        defer klog.InfoS("Shutting down external resizer", "controller", ctrl.name)
 
        if !ctrl.init(ctx) {
                return
        }
 
        stopCh := ctx.Done()
 
        // Starts go-routine that deletes expired slowSet entries.
        go ctrl.slowSet.Run(stopCh)
 
+       wg := sync.WaitGroup{}
        if utilfeature.DefaultFeatureGate.Enabled(features.ReleaseLeaderElectionOnExit) {
                for range workers {
                        wg.Add(1)
                        go func() {
                                defer wg.Done()
                                wait.Until(ctrl.sync, 0, stopCh)
                        }()
                }
        } else {
                for range workers {
                        go wait.Until(ctrl.sync, 0, stopCh)
                }
        }
 
        <-stopCh
+       wg.Wait()
 }
```
That said, let's just pass it a real waitGroup.

pkg/modifycontroller/modify_volume.go
huww98: Refactored these branches. Hopes will make this more readable.

gnufied: I do not see that handled. My concern is - we are reading from `status.TargetVolumeAttributesClassName` and there is a very very small chance that `status` itself may be nil. We do not want external-resizer to crash if that happens.

pkg/modifycontroller/modify_status.go
```diff
@@ -38,69 +37,45 @@ func (ctrl *modifyController) markControllerModifyVolumeStatus(
...
-	// Remove this PVC from the uncertain cache since the status is known now
-	if modifyVolumeStatus == v1.PersistentVolumeClaimModifyVolumeInfeasible {
-		pvcKey, err := cache.MetaNamespaceKeyFunc(pvc)
+		newConditions := make([]v1.PersistentVolumeClaimCondition, 0, 2)
+		newConditions = append(newConditions, v1.PersistentVolumeClaimCondition{
+			Type:               v1.PersistentVolumeClaimVolumeModifyingVolume,
+			Status:             v1.ConditionTrue,
+			LastTransitionTime: metav1.Now(),
+			Message:            conditionMessage,
+		})
```
gnufied, line 53: I think there is a bug here. 

1. I assigned a VAC to the PVC and PVC had `modifyVolume` condition set initially.
2. modification failed, then it will replace both `modifyvolume` and `modifyvolumeerror` conditions:

```
  31   │   conditions:
  32   │   - lastProbeTime: null
  33   │     lastTransitionTime: "2025-09-17T18:48:42Z"
  34   │     message: 'ModifyVolume failed with error: rpc error: code = DeadlineExceeded desc
  35   │       = Taking too long. Waiting for retry.'
  36   │     status: "True"
  37   │     type: ModifyingVolume
  38   │   - lastProbeTime: null
  39   │     lastTransitionTime: "2025-09-17T18:48:42Z"
  40   │     message: 'ModifyVolume failed with error: rpc error: code = DeadlineExceeded desc
  41   │       = Taking too long. Waiting for retry.'
  42   │     status: "True"
  43   │     type: ModifyVolumeError
```

gnufied:
When I apply a invalid `VAC` and then revert back to empty/nil VAC the `status.modifyStatus` is not cleared. IMO it should be cleared:

```
  30   │   modifyVolumeStatus:
  31   │     status: Pending
  32   │     targetVolumeAttributesClassName: foobar
```

pkg/modifycontroller/modify_volume.go
```diff
@@ -53,59 +52,84 @@ func (ctrl *modifyController) modify(pvc *v1.PersistentVolumeClaim, pv *v1.Persi
...
+		return pvc, pv, nil, false
+	}
+
+	// Last modification failed with non-infeasible error
+	inProgress := status != nil && status.Status == v1.PersistentVolumeClaimModifyVolumeInProgress
+
+	if pvcSpecVacName == "" && !inProgress {
+		// User don't care the target state, and we've reached a relatively stable state. Just keep it here.
+		// Note: APIServer generally not allowing setting pvcSpecVacName to empty when curVacName is not empty.
+		klog.V(4).InfoS("stop reconcile for rolled back PVC", "PV", klog.KObj(pv))
```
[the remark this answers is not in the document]
gnufied: >Yes. I think we can clear the conditions. But status.modifyStatus should not be cleared.

That is fine. I was mainly thinking about conditions.

pkg/modifycontroller/modify_status.go
huww98: OK, that makes sense. So we will consistently only report error in `ModifyVolumeError` condition. And make `ModifyVolume operation in progress. ` the only possible message for `ModifyingVolume` condition.

gnufied: I do not think `ModifyVolumeError` should replace `ModifyingVolume` condition.  Both should exist as we discussed.  That is how conditions are typically implemented.

huww98: Implemented, PTAL again. I will test this against our driver too.

huww98:
@gnufied I think we have reached agreement about this: the status.modifyStatus should not be cleared

>> Yes. I think we can clear the conditions. But status.modifyStatus should not be cleared.
>
> That is fine. I was mainly thinking about conditions.

pkg/modifycontroller/modify_status.go
gnufied: This still looks weird IMO. The message associated with `ModifyingVolume` condition should not change on failure. There is no need to complicate this. 

Only `ModifyVolumeError` should have full error message. And currently the `message` field in `ModifyVolumeError` appears truncated. It says - `message: 'invalid parameters: [foobar]'`

huww98:
It is a bit weird if we left `TargetVolumeAttributesClassName` empty while setting `Status` to "Pending". We want them to be set or cleared at the same time since we put them in a struct, right? But as you wish, I reverted to the previous behavior, it is more correct in terms of quota.

I'm thinking that can we just deprecate the Pending status? Maybe we should left `modifyVolumeStatus` untouched if VAC not exists, and just communicate this situation with condition and event.

PR #521, opened 2025-08-28
Requeue over PVC becoming bound

### LINKED ISSUE #520: Resizing PVC close to creation time may result in dropped PVC sync
In our automated environments, we see a rare flake when issuing a resize close to PVC creation time;
The sync loop is short circuited over the PV not existing yet https://github.com/kubernetes-csi/external-resizer/blob/f28f38ba2a7a3375c279d041973ddcb1d76e9bc4/pkg/controller/controller.go#L359-L362 But nothing will requeue over the PV becoming available, so the request would only get processed after `resyncPeriod`

akalenyu:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

There's no logic to requeue over a PVC getting bound,
which means that if we return over no PV existing,
the resize could take up to resyncPeriod.

Fixes #520 

```release-note
BugFix: rare flake when resizing close to creation time (no requeue over "PV bound to PVC not found")
```


gnufied:
So, what if we do not return `nil` when we can't find PV and instead return an error and let pvc get re-queued using rate limited queue? Does this fixes the problem?

I would like to have one object as source of events in the control-loop if we can help. Having both PV and PVC as source of events in control-loop causes all sort of confusion when debugging things.

akalenyu:
> > > That should work, and I was tempted to go there, but the flipside is potentially overwhelming the worker threads over waiting for a PV to become available
> > 
> > 
> > It is rate limited, so should be fine. It is possible that - user has removed the PV while PVC is still there (but they have to remove the finalizer manually from PV etc), but IMO it should be kinda rare. I would rather just return error tbh if PV can't be found and let controller re-queue the request.
> 
> Ok, works for me. I will push the change

Done

PR #573, opened 2026-04-11
Add --node-deployment flag for distributed resizing

### LINKED ISSUE #142: distributed resizing

mnencia:
**What type of PR is this?**

/kind feature

For CSI drivers deployed as DaemonSets managing node-local volumes, there is no central controller pod where a centralized external-resizer can run. The provisioner and snapshotter sidecars already support `--node-deployment` for this deployment model.

This adds the same capability to the external-resizer. When enabled, each instance only processes resize requests for PVs whose nodeAffinity matches the local node. The flag requires `NODE_NAME` env var and is mutually exclusive with `--leader-election`.

The filtering is applied in the controller (`pvNeedResize` and `pvCanBeExpanded`), covering both Resizer implementations without changes to the Resizer interface.

Fixes #142

This is a fresh implementation building on the prior work by @travisghansen in #195 (closed by triage bot), with two key differences:

1. **Filtering location**: applied in the controller rather than in `CanSupport`, so it covers both the `csiResizer` and `trivialResizer` code paths.
2. **Filtering mechanism**: uses PV `nodeAffinity` instead of the PVC `volume.kubernetes.io/selected-node` annotation, which is cleared by the provisioner after binding and would not be available at resize time.

The `pvBelongsToNode` helper checks `MatchExpressions` with `NodeSelectorOpIn` operator, which is what CSI provisioners use when translating `AccessibleTopology` segments into PV node affinity. Other operators (`NotIn`, `Exists`, etc.) and `MatchFields` are intentionally not handled, as they are not used by CSI drivers for node-local topology.

The modify controller (`VolumeAttributesClass`) is not covered by node filtering in this PR. Node-local CSI drivers typically do not use `VolumeAttributesClass`, so this is not blocking for the primary use case and can be addressed in a follow-up.

See also kubernetes-csi/csi-driver-host-path#651 for the downstream consumer of this feature.

```release-note
Add --node-deployment flag to enable distributed resizing for CSI drivers managing node-local volumes in DaemonSet deployments.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ Note that the external-resizer does not scale with more replicas. Only one exter
 
 * `--leader-election-retry-period <duration>`: Duration, in seconds, the LeaderElector clients should wait between tries of actions. Defaults to 5 seconds.
 
+* `--node-deployment`: Enables deploying the external-resizer together with a CSI driver on nodes to manage node-local volumes. When enabled, each resizer instance only handles resize operations for PVs whose node affinity matches the local node. Requires the `NODE_NAME` environment variable to be set and is mutually exclusive with `--leader-election`. Off by default.
+
 * `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `ControllerExpandVolume` calls. 10 seconds is used by default.
 
 * `-kube-api-burst <int>` : Burst to use while communicating with the kubernetes apiserver. Defaults to 10. (default 10).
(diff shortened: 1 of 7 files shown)


mnencia:
I have a working integration of this at [`dev/651-with-resizer`](https://github.com/mnencia/csi-driver-host-path/tree/dev/651-with-resizer) on csi-driver-host-path, adding the csi-resizer with `--node-deployment=true` to the distributed DaemonSet deployment. Tested on a 3-node Kind cluster with volume expansion working correctly (including the node-filtering logic skipping non-local PVs as expected). That branch depends on this PR being merged. Related to #142.

PR #581, opened 2026-05-20
Move feature to that restores pv size as annotation beta

gnufied:
This feature already is supported in `k/k` and requires no featuregate.  

It allows external-resizer to continue resizing volumes if PVC was deleted while resizing was pending on PV. 

```release-note
Allow resuming of resize operation if PVC is deleted while resize is pending
```

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -29,6 +29,7 @@ import (
 const (
 	// owner: @sunpa93
 	// alpha: v1.22
+	// beta: v2.2
 	AnnotateFsResize featuregate.Feature = "AnnotateFsResize"
 
 	// owner: @gnufied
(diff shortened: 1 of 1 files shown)


xing-yang:
What is the expected end results of this feature?  The PV will be resized successfully but not bound to any PV?  Is there a KEP for this feature?

gnufied:
I added hold previously to verify if restore works correctly with new code and it does.

/hold cancel

gnufied:
@xing-yang this fixes a bug opened by you actually - https://github.com/kubernetes/kubernetes/issues/88683

torredil (commented):
Besides the pending e2e test this lgtm.

gnufied:
I have added new e2e, but the problem is - they won't start passing until this feature is enabled in image-build they are using https://github.com/kubernetes/kubernetes/pull/139203 

Another thing I can do is - I can explicitly enable the featuregate is CI, to make it all green for now I guess.

gnufied:
I have explicitly enabled the featuregate in e2e tests - https://github.com/kubernetes/kubernetes/pull/139203 and it is all green now.

xing-yang:
e2e passed.

/lgtm

PR #584, opened 2026-05-27
Update changelog and readme for 2.2 release

gnufied:
/release-note-none

### DIFF
--- a/CHANGELOG/CHANGELOG-2.2.md
+++ b/CHANGELOG/CHANGELOG-2.2.md
@@ -1,0 +1,80 @@
+## Release notes for 2.2.0
+
+[Documentation](https://kubernetes-csi.github.io)
+
+
+# Changelog since 2.1.0
+
+## Changes by Kind
+
+### Other (Cleanup or Flake)
+
+- Bump k8s dependencies to v1.36.1 (#578, @dfajmon)
+
+### Feature
+
+- Allow resuming of resize operation if PVC is deleted while resize is pending (#581, @gnufied)
+- Move RecoverVolumeExpansionFailure to GA (#577, @gnufied)
+
+## Dependencies
+
+### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG/CHANGELOG-2.2.md
```diff
@@ -0,0 +1,80 @@
...
+[Documentation](https://kubernetes-csi.github.io)
+
+
+# Changelog since 2.1.0
+
+## Changes by Kind
+
+### Other (Cleanup or Flake)
+
+- Bump k8s dependencies to v1.36.1 (#578, @dfajmon)
```
xing-yang: Can you use hyperlinks like this? 
([#1363](https://github.com/kubernetes-csi/external-snapshotter/pull/1363), [@dfajmon](https://github.com/dfajmon))

CHANGELOG/CHANGELOG-2.2.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Feature
+
+- Allow resuming of resize operation if PVC is deleted while resize is pending (#581, @gnufied)
+- Move RecoverVolumeExpansionFailure to GA (#577, @gnufied)
```
xing-yang: I'm wondering if we should bump the major version now that this feature is GA.

jsafrane: Do CSI driver vendors do anything (change RBAC, change the resizer cmdline, ...)? If not, then we probably don't need a new major version.

gnufied: The CSI driver vendors don't need to change anything. It is basically same as beta feature. I tend to agree that we don't need a major version bump for this.

xing-yang: Ok.

CHANGELOG/CHANGELOG-2.2.md
gnufied: fixed.

PR #596, opened 2026-08-05
Add separate --resize-timeout and --modify-timeout flags

xing-yang:
**What type of PR is this?**
/kind feature

`--timeout` is currently used as the per-call timeout for both `ControllerExpandVolume`
and `ControllerModifyVolume`, with no way to configure them independently. This is a
problem for CSI drivers where a modify-volume operation can
potentially take much longer than a resize call, but operators have no way to give
modify volume a longer timeout without also inflating the resize timeout (or vice versa).

This PR adds two new flags, `--resize-timeout` and `--modify-timeout`, so each operation's
timeout can be set independently. Both default to the same value as `--timeout` (10s), so
behavior is unchanged for anyone not using the new flags.

For backwards compatibility, `--timeout` still works exactly as before: if it is explicitly
set on the command line, it overrides both `--resize-timeout` and `--modify-timeout`. If
`--timeout` is left at its default, `--resize-timeout`/`--modify-timeout` take effect
independently. `--timeout` is unchanged for its other use (CSI driver socket / GetPluginInfo).

Fixes #

```release-note
Add separate --resize-timeout and --modify-timeout flags
```

### DIFF
--- a/cmd/csi-resizer/main.go
+++ b/cmd/csi-resizer/main.go
@@ -68,7 +68,14 @@ var (
 
 	extraModifyMetadata = flag.Bool("extra-modify-metadata", false, "If set, add pv/pvc metadata to plugin modify requests as parameters.")
 
-	timeout = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for CSI driver socket.")
+	timeout = flag.Duration("timeout", 10*time.Second, "Timeout for waiting for CSI driver socket. "+
+		"If set explicitly on the command line, this value also overrides --resize-timeout and "+
+		"--modify-timeout as the per-call timeout for ControllerExpandVolume and ControllerModifyVolume.")
+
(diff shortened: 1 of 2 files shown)


cmd/csi-resizer/main.go
```diff
@@ -80,6 +87,25 @@ var (
...
+		}
+	})
+	if timeoutExplicit {
+		return timeout, timeout
+	}
+	return resizeTimeout, modifyTimeout
+}
+
 func main() {
 	flag.Var(cflag.NewMapStringBool(&featureGates), "feature-gates", "A set of key=value paris that describe feature gates for alpha/experimental features for csi external resizer."+"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
```
sunnylovestiramisu, line 110: oops not related to this PR but i think this should be "pairs"

sunnylovestiramisu:
We can add some checks like timeouts should be >= a safe guard number etc. But that is optional, the previous timeout does not have that check anyway.

/lgtm

